PR #25, opened 2017-06-11
Removed unnecessary Silverlight preprocessor directives and classes

gpetrou:


Tests/RowanTest.Common/Runner.cs
```diff
@@ -80,10 +78,6 @@ public static class Runner {
         static Runner() {
             if (System.IntPtr.Size == 8) _Is64 = true;
 
-#if SILVERLIGHT
-            _IsSilverlight = true;
```
slozier: Maybe get rid of the variable definition as well?

gpetrou: Sure.

gpetrou:
OK, but in that case are you squashing the commits when you merge?
Personally, I prefer having one commit per PR.

slozier:
Yes, normally we squash and merge.

PR #26, opened 2017-06-14
Convert to auto-property

gpetrou:


Src/Microsoft.Dynamic/Runtime/CompilerContext.cs
```diff
@@ -22,46 +22,13 @@ namespace Microsoft.Scripting.Runtime {
     /// from this class to provide additional contextual information.
     /// </summary>
     public sealed class CompilerContext {
+        public SourceUnit SourceUnit { get; }
 
-        /// <summary>
-        /// Source unit currently being compiled in the CompilerContext
```
slozier: Can we move these comments to the properties?

Src/Microsoft.Dynamic/zlib.net/ZStream.cs
```diff
@@ -121,40 +121,22 @@ public sealed class ZStream
...
-        /// </summary>
-        private BlockType data_type;
-		
-        /// <summary>
-        /// A checksum computed with Adler algorithm
-        /// </summary>
-        private long _adler;
 
-        #endregion
+	    #endregion
```
slozier: Is there an indentation issue here? (Same thing for the summaries below).

Src/Microsoft.Dynamic/Runtime/CompilerContext.cs
gpetrou: Sure. Should we follow StyleCop rules for such comments? "Gets or sets etc..." http://stylecop.soyuz5.com/SA1623.html

slozier: We probably should, but I expect most of the codebase doesn't so I'm not too worried about it.

Src/Microsoft.Dynamic/zlib.net/ZStream.cs
gpetrou: Apparently there are some tabs from previous code. I will replace some of them with whitespace.

gpetrou:
Can we merge this?

PR #36, opened 2017-06-28
Upgrade to VS2017 and new csproj format

gpetrou:


Tests/Metadata/Metadata.csproj
```diff
@@ -12,39 +12,8 @@
...
-    <ProjectReference Include="..\..\Src\Microsoft.Dynamic\Microsoft.Dynamic.csproj">
-      <Project>{EB66B766-6354-4208-A3D4-AACBDCB5C3B3}</Project>
-      <Name>Microsoft.Dynamic</Name>
-    </ProjectReference>
-    <ProjectReference Include="..\..\Src\Microsoft.Scripting.Metadata\Microsoft.Scripting.Metadata.csproj">
-      <Project>{ACDD9B9E-8FE6-439C-9521-1CCBA47F6143}</Project>
-      <Name>Microsoft.Scripting.Metadata</Name>
-    </ProjectReference>
+    <ProjectReference Include="..\..\Src\Microsoft.Scripting.Metadata\Microsoft.Scripting.Metadata.csproj" />
+    <ProjectReference Include="..\..\Src\Microsoft.Scripting\Microsoft.Scripting.csproj" />
```
slozier: This one should point to Microsoft.Dynamic.

gpetrou: Ooops, yes, I forgot to revert this. The problem I have is something else though with the ReflectionUtils.

slozier: In the AssemblyInfo files for Microsoft.Scripting and Microsoft.Dynamic there are some `InternalsVisibleTo("TestInternalDLR, ...")` which may be what's causing the issues. You can comment those out.

Src/Microsoft.Scripting/Microsoft.Scripting.csproj
```diff
@@ -11,133 +11,9 @@
...
-  </PropertyGroup>
-  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'v4Release|AnyCPU' ">
-  </PropertyGroup>
-  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
-  </PropertyGroup>
-  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
-  </PropertyGroup>
+  </PropertyGroup>  
   <ItemGroup>
-    <Compile Include="..\AssemblyInfo.cs">
```
slozier: I think we'll want to keep this one (here and also in the other csproj files). I probably inadvertently removed it from my original PR. Note that you make have to change the Link to something else (e.g. Properties\AssemblyInfo.Generated.cs) if it causes conflicts with an existing AssemblyInfo.cs file.

Build/Common.proj
```diff
@@ -49,7 +55,6 @@
     <BuildFlavour>Debug</BuildFlavour>
     <BuildFlavour Condition="'$(Configuration)' == 'FxCop'">Release</BuildFlavour>
     <BuildFlavour Condition="'$(Configuration)' == 'Release'">Release</BuildFlavour>
-    <BuildFlavour Condition="'$(Configuration)' == 'v4Release'">Release</BuildFlavour>
```
slide: Why was this removed?

Tests/Metadata/Metadata.csproj
slide: Those are needed for tests though, I believe.

slozier: You're right, although it looks like we only use the internals of Microsoft.Scripting so we can exclude it from Microsoft.Dynamic.

slide:
It would be nice if .travis.yml would use the makefile instead.

slide:
Does this change mean that you MUST have VS2017 in order to build?

slozier:
If you want to use VS then yes, it won't work with VS2015.

slide:
I'm not sure I have access to VS2017

gpetrou:
If I am not wrong the new csproj format is compatible only with VS2017.
So, yes, you will need to get VS2017.
I am not sure about the makefile in travis. Maybe this should be a separate issue/PR?

slozier:
Looks like there's still some stuff ending up in the bin/Release folder. For example Microsoft.Scripting.AspNet which I hadn't converted to the new csproj because I have no idea what it's used for.

slide:
It's used for doing ASP.NET Webform (I believe) in IronPython. I think we can probably get rid of it since Webforms is not really used much anymore.

gpetrou:
I was planning to add a separate PR for the rest of the cspoj files.
Some of them don't even seem to be part of the sln.

Tests/ClrAssembly/ClrAssembly.csproj
```diff
@@ -1,25 +1,20 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="4.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+﻿<Project Sdk="Microsoft.NET.Sdk">
   <Import Project="$(ProjectDir)..\..\Build\Common.proj" />
   <PropertyGroup>
-    <SchemaVersion>2.0</SchemaVersion>
+    <TargetFrameworks>net40;net45</TargetFrameworks>
+    <PlatformTarget>anycpu</PlatformTarget>
```
slozier: Can you get rid of all those extra properties that were added?

slozier:
The reason it's failing on Travis is that it's using Build.proj which will try to build Release and v4Release (and v4Release doesn't exist anymore). If you remove the `<ReferencedPlatform Include="net40">...` it should work. We'll probably need to do some more cleanup of Build.proj.

Tests/ClrAssembly/ClrAssembly.csproj
gpetrou: Sorry, I am still experimenting please ignore for now.

slozier:
You can add `<EnableDefaultCompileItems>false</EnableDefaultCompileItems>` to ClrAssembly.csproj to resolve the compile issues (you'll need to re-add dummy.cs to get rid of the warning).

gpetrou:
Indeed, there is some information in https://docs.microsoft.com/en-us/dotnet/core/tools/csproj

Src/Microsoft.Dynamic/Microsoft.Dynamic.csproj
```diff
@@ -12,433 +12,16 @@
...
-    <Reference Include="System.Configuration" />
-    <Reference Include="System.Runtime.Remoting" />
-  </ItemGroup>
-  <ItemGroup Condition="'$(ReferencedPlatform)' == 'net40'">
-    <Reference Include="System.Xaml" />
-  </ItemGroup>
-  <ItemGroup Condition="'$(ReferencedPlatform)' == 'net45'">
-    <Reference Include="System.Xaml" />
-  </ItemGroup>
+    <ProjectReference Include="..\Microsoft.Scripting.Metadata\Microsoft.Scripting.Metadata.csproj" />
```
slozier: Can you add a `Condition=" $(Features.Contains('FEATURE_METADATA_READER')) "` to this reference?

slozier:
I tried building locally and it looks like ClrAssembly no longer produces the test DLLs?

Tests/ClrAssembly/ClrAssembly.csproj
```diff
@@ -99,15 +93,11 @@
   </Target>
   <Target Name="BeforeBuild" DependsOnTargets="LoadOrder;ReachType;CSManyArea;VBManyArea" />
   <ItemGroup>
-    <Compile Include="dummy.cs" />
```
slozier: Can you keep this one in the project (to eliminate the CS2008 build warning).

Src/Microsoft.Dynamic/Microsoft.Dynamic.csproj
gpetrou: Yes, sure. Is it causing a problem?

slozier: Not at the moment, but it's only used when that condition is met and will need to be excluded for .NET Core.

Src/Microsoft.Dynamic/Microsoft.Dynamic.csproj
```diff
@@ -23,5 +23,9 @@
     <Reference Include="System.Configuration" Condition=" $(Features.Contains('FEATURE_CONFIGURATION')) " />
     <Reference Include="System.Runtime.Remoting" Condition=" $(Features.Contains('FEATURE_REMOTING')) " />
     <Reference Include="System.Xaml" Condition=" $(Features.Contains('FEATURE_WPF')) " />
-   </ItemGroup>
+  </ItemGroup>
+  <Import Project="$(BeforeTargetFiles)" />
+  <Target Name="BeforeBuild" DependsOnTargets="$(BeforeTargets);GenerateCurrentVersion" />
```
slozier: I think these won't work properly (same issue as before, the target name BeforeBuild doesn't seem to work with the new project files).

Src/Microsoft.Dynamic/Microsoft.Dynamic.csproj
```diff
@@ -24,8 +24,4 @@
     <Reference Include="System.Runtime.Remoting" Condition=" $(Features.Contains('FEATURE_REMOTING')) " />
     <Reference Include="System.Xaml" Condition=" $(Features.Contains('FEATURE_WPF')) " />
   </ItemGroup>
-  <Import Project="$(BeforeTargetFiles)" />
-  <Target Name="BeforeBuild" DependsOnTargets="$(BeforeTargets);GenerateCurrentVersion" />
```
slozier: We're still going to need these. We just have to change them to something like `<Target Name="BeforeBuildStart" BeforeTargets="BeforeBuild"`.

gpetrou: Hmm, I did the change but still get some errors on a clean build.

slozier:
Looks like it's still not generating the CurrentVersion files. We'll probably have to tweak it some more to get the target to run. I'm gonna have to do some reading on the subject...

gpetrou:
I tried to do it like https://stackoverflow.com/questions/43921992/how-can-i-use-beforebuild-and-afterbuild-targets-with-visual-studio-2017 but no luck yet.

slozier:
I ran a few tests. Looks like the files are getting generated they're just not picked up automatically. Maybe BeforeTargets needs to be something other than BeforeBuild? Maybe there's a target that occurs before file detection?

I found a workaround but it's ugly so I'm open to suggestions. I added the following after the BeforeBuildStarts targets:
```
  <ItemGroup>
    <!-- if the file does not exist it's not picked up automatically -->
    <Compile Remove="Properties\CurrentVersion.Generated.cs" />
    <Compile Include="Properties\CurrentVersion.Generated.cs" />
  </ItemGroup>
```

gpetrou:
I had a look around, but could not find another solution, so I used your suggestion for now.

Src/Microsoft.Scripting/Microsoft.Scripting.csproj
```diff
@@ -1,8 +1,9 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="4.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+﻿<Project>
+  <Import Project="Sdk.props" Sdk="Microsoft.NET.Sdk" />
```
slozier: You're doing the separate target/props import instead of using sdk in the project tag so that the BeforeBuild/AfterBuild targets work? Perhaps a comment to that effect in the project files would be useful.

gpetrou:
I can do a PR for the rest of the csproj files once this is merged.
Would it be better to convert the make.cmd to a PowerShell script?
Personally, I find these files harder to work with unless they are doing very simple tasks.

slide:
I would prefer something like fake or psake if possible.

slide:
Actually, I would prefer we move everything to msbuild so we could use the same commands on both windows and Linux, e.g., msbuild /t:test-all

PR #39, opened 2017-07-02
Remove redundant qualifiers

gpetrou:


slide:
My concern with some of these is that there is very little localized testing in the DLR repo. Things might start popping up once we start running this with IronPython. Is there anything we can do to run some tests with IronPython as well?

PR #42, opened 2017-07-02
Use string interpolation

gpetrou:


Src/Microsoft.Dynamic/Hosting/Shell/Remote/RemoteConsoleCommandLine.cs
```diff
@@ -45,7 +45,8 @@ public class RemoteConsoleCommandLine : CommandLine {
                 // All bets are off while accessing the remote runtime. So we catch all exceptions.
                 // However, in most cases, we only expect to see RemotingException here.
                 if (!(exceptionDuringHandling is RemotingException)) {
-                    Console.WriteLine(String.Format("({0} thrown while trying to display unhandled exception)", exceptionDuringHandling.GetType()), Style.Error);
```
slide: This one could have just removed the String.Format and used the WriteLine overload which takes params.

Src/Microsoft.Scripting/Runtime/DynamicStackFrame.cs
```diff
@@ -54,14 +54,10 @@ public class DynamicStackFrame {
...
-        public override string ToString() {
-            return string.Format(
-                "{0} in {1}:{2}, {3}",
-                _funcName ?? "<function unknown>",
-                _filename ?? "<filename unknown>",
-                _lineNo,
-                (_method != null ? _method.ToString() : "<method unknown>")
-            );
+        public override string ToString()
+        {
```
slide: Nitpick- brace should be on same line as method name

Src/Microsoft.Dynamic/Hosting/Shell/Remote/RemoteConsoleCommandLine.cs
gpetrou: Style.Error is still there. Are you talking about something else?

slozier: He probably though this was System.Console.WriteLine.

slide: Ah, you are correct. Apparently I need to look at filenames :-)

PR #43, opened 2017-07-02
Merge sequential checks

gpetrou:


Src/Microsoft.Dynamic/zlib.net/Inflate.cs
```diff
@@ -259,12 +259,12 @@ internal int inflateInit(ZStream z, int windowBits)
 		internal int inflate(ZStream z, FlushStrategy flush)
 		{
 		    int internalFlush = (int)flush;
-
-            int res_temp;
 			
-			if (z == null || z.istate == null || z.next_in == null)
+			if (z?.istate == null || z.next_in == null)
```
slozier: I'd prefer keeping the null check separate as it's clearer that it's also not null in the second condition.

Src/Microsoft.Dynamic/zlib.net/Inflate.cs
```diff
@@ -487,7 +487,7 @@ internal int inflateSetDictionary(ZStream z, byte[] dictionary, int dictLength)
 		{
 			int index = 0;
 			int length = dictLength;
-			if (z == null || z.istate == null || z.istate.mode !=  InflateMode.DICT0)
+			if (z?.istate == null || z.istate.mode !=  InflateMode.DICT0)
```
slozier: Does `z?.istate?.mode != ...` work?

gpetrou:
I also replaced some tabs with spaces. Seems like some of these files have mixed tabs with spaces.

Src/Microsoft.Dynamic/zlib.net/Inflate.cs
```diff
@@ -487,7 +487,7 @@ internal int inflateSetDictionary(ZStream z, byte[] dictionary, int dictLength)
 		{
 			int index = 0;
 			int length = dictLength;
-			if (z == null || z.istate == null || z.istate.mode !=  InflateMode.DICT0)
+			if (z?.istate?.mode != InflateMode.DICT0)
```
slide: Doesn't this change the meaning of the if check?

slozier: If z or z.istate are null then z?.istate?.mode is null so it's != InflateMode.DICT0. I agree it's not the most obvious conversion.

slide: All these changes make me want to go back in my SW at work and actually use new features of C# :-)

PR #44, opened 2017-07-02
Use null propagation

gpetrou:


Src/Microsoft.Dynamic/Generation/MethodSignatureInfo.cs
```diff
@@ -37,9 +37,8 @@ public MethodSignatureInfo(MethodInfo info)
 
         public override bool Equals(object obj) {
             MethodSignatureInfo args = obj as MethodSignatureInfo;
-            if (args == null) return false;
 
-            if (args._isStatic != _isStatic || args._pis.Length != _pis.Length || args._genericArity != _genericArity) {
+            if (args?._isStatic != _isStatic || args._pis.Length != _pis.Length || args._genericArity != _genericArity) {
```
slozier: Same as before (not a fan) although I find this case more confusing since it's an or.

Src/Microsoft.Dynamic/Actions/ComboBinder.cs
```diff
@@ -242,31 +242,25 @@ public class ParameterMappingInfo {
 
         public override bool Equals(object obj) {
             ParameterMappingInfo pmi = obj as ParameterMappingInfo;
-            if (pmi == null) {
-                return false;
-            }
-
-            if (pmi.ParameterIndex == ParameterIndex && pmi.ActionIndex == ActionIndex) {
+            if (pmi?.ParameterIndex == ParameterIndex && pmi.ActionIndex == ActionIndex) {
```
slozier: I'm not a fan of using the operator in this type of situation (the possibly null value is used in more than one of the conditions).

Src/Microsoft.Dynamic/Actions/Calls/TypeInferer.cs
```diff
@@ -414,21 +414,19 @@ class ArgumentInputs {
             if (parameterType.IsSubclassOf(typeof(Delegate))) {
                 // see if we have an invokable object which can be used to infer into this delegate
                 IInferableInvokable invokeInfer = input as IInferableInvokable;
-                if (invokeInfer != null) {
```
slozier: Seems like this one is better suited for pattern matching. Assuming we can use it with Mono...

Src/Microsoft.Dynamic/Interpreter/Instructions/InstructionList.cs
```diff
@@ -372,11 +372,9 @@ internal struct InstructionView {
         internal void SwitchToBoxed(int index, int instructionIndex) {
             var instruction = _instructions[instructionIndex] as IBoxableInstruction;
 
-            if (instruction != null) {
-                var newInstruction = instruction.BoxIfIndexMatches(index);
-                if (newInstruction != null) {
-                    _instructions[instructionIndex] = newInstruction;
-                }
+            var newInstruction = instruction?.BoxIfIndexMatches(index);
```
slozier: Pattern matching instead?

slide:
This now has conflicts, also please check on the suggestions from @slozier

Src/Microsoft.Dynamic/Actions/Calls/TypeInferer.cs
gpetrou: Hmm, not sure about the pattern matching suggestions, since these are nested checks. Can you provide an example?

slozier: `if (input is IInferableInvokable invokeInfer) {` instead of as followed by a null conditional.

gpetrou:
Can we merge this now?

PR #47, opened 2017-07-02
Fix some CA1709 warnings

gpetrou:


slozier:
I'm not a big fan of the CA1709 warning. I guess in this case it's fine since it's limited to zlib.

Src/Microsoft.Dynamic/zlib.net/Deflate.cs
```diff
@@ -2061,8 +2061,8 @@ internal int deflate(ZStream strm, FlushStrategy f)
 								head[i] = 0;
 						}
 					}
-					strm.flush_pending();
-					if (strm.avail_out == 0)
+					strm.FlushPending();
+                    if (strm.avail_out == 0)
```
slide: whitespace change?

gpetrou: I think this is because some of these files have mixed tabs and spaces. My changes always use spaces.

PR #48, opened 2017-07-06
Workaround for IronLanguages/ironpython2#27

slide:
Workaround for IronLanguages/ironpython2#27


slide:
The testcase from the issue on Bugzilla prints the correct stack trace. Not a full test case for sure. I'm hoping to get some more feedback on the bugzilla.

PR #50, opened 2017-07-07
Remove FEATURE_CORE_DLR

slozier:


Src/Microsoft.Scripting/Runtime/LanguageContext.cs
```diff
@@ -13,16 +13,10 @@
  *
  * ***************************************************************************/
 
-#if FEATURE_CORE_DLR
 using System.Linq.Expressions;
 #if !WIN8
```
slide: We can remove WIN8 too I think, but we can make it another PR if desired.

slozier: I'll leave it for a separate PR.

PR #51, opened 2017-07-07
Add some MathUtils tests

gpetrou:
I suspect I need to check the appveyor/travis files to run these?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,9 @@ bin/
 /Package
 /Util/References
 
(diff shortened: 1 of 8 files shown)


slide:
appveyor will supposedly autorun tests if you have something in your appveyor.yml, but doing something explicit would be good so it can be used for both.

slozier:
I'm not seeing the Appveyor results. Did we break something?

PR #60, opened 2017-07-12
Convert to auto-property and expression-bodied property

gpetrou:


gpetrou:
Can we merge this now?

PR #61, opened 2017-07-12
Fix documentation files path

gpetrou:


slozier:
Do we need the leading slashes?

gpetrou:
Locally, yes but it seems we still need full path the CI builds, so I changed the parameter.

slozier:
Hmm, I tried over here with just `$(AssemblyName).xml` and it worked fine.

gpetrou:
In my case it creates the xml inside the project folder.

slozier:
Odd. I wonder what the difference is...

PR #62, opened 2017-07-12
Add restore before build

slozier:


slozier:
Bah, looks like this may not work in Travis since it requires msbuild 15.1+...

PR #64, opened 2017-07-17
Join declaration and assignment

gpetrou:


Src/Microsoft.Scripting/Runtime/DynamicOperations.cs
```diff
@@ -155,8 +155,7 @@ public sealed partial class DynamicOperations {
         /// Gets the member name from the object obj.  Throws an exception if the member does not exist or is write-only.
         /// </summary>
         public object GetMember(object obj, string name, bool ignoreCase) {
-            CallSite<Func<CallSite, object, object>> site;
-            site = GetOrCreateSite<object, object>(_lc.CreateGetMemberBinder(name, ignoreCase));
+            CallSite<Func<CallSite, object, object>> site = GetOrCreateSite<object, object>(_lc.CreateGetMemberBinder(name, ignoreCase));
```
slide: I would prefer var in this case, if it can be used.

Src/Microsoft.Scripting/Runtime/DynamicOperations.cs
```diff
@@ -351,8 +350,7 @@ public sealed partial class DynamicOperations {
         /// Converts the object obj to the type type including implicit conversions.
         /// </summary>
         public object ImplicitConvertTo(object obj, Type type) {
-            CallSite<Func<CallSite, object, object>> site;
-            site = GetOrCreateSite<object, object>(_lc.CreateConvertBinder(type, false));
+            CallSite<Func<CallSite, object, object>> site = GetOrCreateSite<object, object>(_lc.CreateConvertBinder(type, false));
```
slide: Again, would prefer var.

Src/Microsoft.Scripting/Runtime/DynamicOperations.cs
gpetrou: Hmm, in these cases it does not seem to be very clear of what the method returns.
Should we leave these as they are?

slozier: Ah, the eternal var debate. I agree that in this case it's not clear what the method returns, but I also agree that it's a long and ugly type. We should probably come up with some guidelines..

slide: Let's leave it as is (not using var).

PR #70, opened 2017-07-20
Add .NET Core support to DLR

slozier:


Build/Common.proj
```diff
@@ -59,8 +59,7 @@
 
   <PropertyGroup Condition="'$(ReferencedPlatform)' == ''">
     <ReferencedPlatform>net45</ReferencedPlatform>
```
slide: Do we need ReferencedPlatform at all anymore?

Build/Common.proj
```diff
@@ -73,7 +72,7 @@
   </PropertyGroup>
 
   <!-- References -->
-  <Import Project="$(DlrBuildDir)\$(ReferencedPlatform).props" />
```
slide: Is TargetFramework always set? If not, don't we need this so that it pulls in when it's not set?

Dlr.sln
```diff
@@ -53,10 +52,6 @@ Global
 		Release|Any CPU = Release|Any CPU
 	EndGlobalSection
 	GlobalSection(ProjectConfigurationPlatforms) = postSolution
-		{3272E46C-F4D9-418D-9EE9-081A1B658EE4}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
```
slide: What project was removed?

Src/Microsoft.Dynamic/Generation/FieldBuilderExpression.cs
```diff
@@ -31,9 +31,19 @@ namespace Microsoft.Scripting.Generation {
     public class FieldBuilderExpression : Expression {
         private readonly FieldBuilder _builder;
 
+#if NETCOREAPP2_0
+        private readonly StrongBox<Type> _finishedType;
```
slide: Can we do this as a runtime check instead, e.g. make finishedType default to null and then check if its null in GetFieldInfo?

Build/Common.proj
slozier: I was planning on eventually getting rid of it, but it's somewhat unrelated to the .NET Core support so I was going to do it as a separate PR.

slide: Ok, as long as it is on the radar.

Build/Common.proj
slozier: With the new csproj files TargetFramework should always be set. I'm not exactly sure how it behaves when you mix with old style csproj files...

Dlr.sln
slozier: Microsoft.Scripting.AspNet. I don't mind re-adding it, but it should be updated to the new csproj format if we do.

slide: Removing that one is fine

Src/Microsoft.Dynamic/Generation/FieldBuilderExpression.cs
slozier: The code wouldn't compile since there's no FieldBuilder.GetToken in .NET Core. I just readded the code that was there before it was removed in the Silverlight cleanup.

slide: :+1:

Build/Common.proj
slide: Why the need for the Condition in the import of $(TargetFramework).props then?

slozier: I'm not super familiar with how it works, but msbuild does an "outer" build and an "inner" build. In the "outer" build pass TargetFramework is not set. You can try to read about it if you want, I don't understand much of it: https://oren.codes/2017/01/04/multi-targeting-the-world-a-single-project-to-rule-them-all/#how-it-works

Src/Microsoft.Dynamic/Microsoft.Dynamic.csproj
```diff
@@ -1,8 +1,8 @@
-﻿<Project>
-  <Import Project="Sdk.props" Sdk="Microsoft.NET.Sdk" />
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
gpetrou: Is there a reason that this changed? Did you find a problem with the current implementation?

Tests/ClrAssembly/ClrAssembly.csproj
```diff
@@ -74,7 +75,11 @@
   </Target>
   <Target Name="CSManyArea" Inputs="@(AreaCSFile)" Outputs="@(AreaCSFile->'$(OutputPath)\$(Prefix)%(Filename).dll')" DependsOnTargets="TypeSample">
     <Csc TargetType="library" Sources="Src\fieldtests.cs" OutputAssembly="$(OutputPath)\$(Prefix)fieldtests.dll" References="$(OutputPath)\$(Prefix)typesamples.dll" />
-    <Csc TargetType="library" Sources="Src\methodargs.cs" OutputAssembly="$(OutputPath)\$(Prefix)methodargs.dll" References="System.dll;$(OutputPath)\Microsoft.Scripting.dll;$(OutputPath)\$(Prefix)typesamples.dll" />
+    <PropertyGroup>
+      <SystemRuntimeDll></SystemRuntimeDll>
```
gpetrou: Is this needed?

slozier: You mean the empty assignment? Not sure, will try.

Src/Microsoft.Dynamic/Microsoft.Dynamic.csproj
slozier: Yes, the current implementation was causing issues (although at this point I don't remember what exactly).

PR #76, opened 2017-08-24
Remove BuildNetCoreApp condition and build .NET Core 2.0 libraries

gpetrou:


.travis.yml
```diff
@@ -16,20 +16,24 @@ solution: Dlr.sln
...
-dotnet: 2.0.0-preview2-006497
+dotnet: 2.0.0
 
 script:
   - dotnet restore Dlr.sln --packages ./packages
   - msbuild Build.proj /t:Build "/p:Mono=true;BuildFlavour=$CONFIGURATION"
   - pushd bin/$CONFIGURATION
   - mono ./net40/TestInternalDLR.exe
   - mono ./net45/TestInternalDLR.exe
+  - mono ./netcoreapp2.0/TestInternalDLR.exe
```
slozier: .NET Core doesn't produce an exe (and in any case we shouldn't be testing the .NET Core stuff with Mono). You could try "dotnet ./netcoreapp2.0/TestInternalDLR.dll".

.travis.yml
```diff
@@ -16,20 +16,24 @@ solution: Dlr.sln
...
   - mono ./net40/Metadata.exe
   - mono ./net45/Metadata.exe
+  - mono ./netcoreapp2.0/Metadata.exe
   - mono ./../../packages/nunit.consolerunner/3.6.1/tools/nunit3-console.exe ./net40/Microsoft.Dynamic.Test.dll
   - mono ./../../packages/nunit.consolerunner/3.6.1/tools/nunit3-console.exe ./net45/Microsoft.Dynamic.Test.dll
+  - mono ./../../packages/nunit.consolerunner/3.6.1/tools/nunit3-console.exe ./netcoreapp2.0/Microsoft.Dynamic.Test.dll
```
slozier: I'm don't think the runner will work with .NET Core. We could convert them to be console apps instead (like IronPythonTest), but I'm also fine with skipping these tests on .NET Core for now.

appveyor.yml
```diff
@@ -76,12 +76,16 @@ build_script:
 test_script:
   - '%APPVEYOR_BUILD_FOLDER%/bin/%CONFIGURATION%/net40/TestInternalDLR.exe'
   - '%APPVEYOR_BUILD_FOLDER%/bin/%CONFIGURATION%/net45/TestInternalDLR.exe'
+  - '%APPVEYOR_BUILD_FOLDER%/bin/%CONFIGURATION%/netcoreapp2.0/TestInternalDLR.exe'
```
slozier: Same comments as with travis.yml...

gpetrou:
When building locally, CodeDom seems fine.
I will give it a try and remove the preview.

.travis.yml
gpetrou: Yes, seems like the NUnit team is still working on this. Maybe we should wait for them, instead of adding console projects for now.

slozier:
Hmm, the AppVeyor may be related to the nuget version? Not sure for the Travis failure...

gpetrou:
Yes, I think I need to edit the AppVeyor file to download a newer Nuget.exe
Also, one of the DLLs is x86, so it gives BadImageFormatException.
Not sure if/how I can invoke an x86 dotnet in the CI.
There is one more problem with TestReflectionServices which I am not sure about.
I will probably check this again during the weekend.

slozier:
x86: I see, there are actually two `PlatformTarget` entries in Metadata.csproj, we should remove the x86 one and keep anycpu.

slozier:
For TestReflectionServices: .NET Core has different assembly layouts than the full framework so that could be the issue. Another thing to try is adding FEATURE_METADATA_READER to netcoreapp2.0.props (since you fixed it to work with .NET Core anyway).

gpetrou:
PlatformTarget x86 was the original one though. AnyCPU was added by mistake in https://github.com/IronLanguages/dlr/pull/36/files#diff-0101e7dd4da19af70160bb466410a2abR5
Perhaps x86 was used on purpose before.

slozier:
I would still drop x86 and change it to anycpu (unless it's causing failures in net40 or net45).

gpetrou:
Trying to use http://www.alteridem.net/2017/05/04/test-net-core-nunit-vs2017/
to run the other tests from command line gives an error about Microsoft.NET.Test.Sdk and net40.
I suspect either NuGet does not like this or there is an actual incompatibility.

slozier:
Seems like the Travis build is not picking up on the .NET Core reference assembly?

Possible solutions:
- Remove the reference to the assembly and see what sort of error it gives with the Mono csc. Maybe it'll give a clue as to how to resolve the issue.
- Use the full path to the assembly. 
- Make a separate project file to generate the methodargs DLL.

Src/Microsoft.Dynamic/Microsoft.Dynamic.csproj
```diff
@@ -1,8 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <Import Project="$(ProjectDir)..\..\Build\Common.proj" />
   <PropertyGroup>
-    <TargetFrameworks>net40;net45</TargetFrameworks>
-    <TargetFrameworks Condition="'$(BuildNetCoreApp)' == 'true'">$(TargetFrameworks);netcoreapp2.0</TargetFrameworks>
+    <TargetFrameworks>net40;net45;netcoreapp2.0</TargetFrameworks>
```
slide: I'd like this to have a condition on it if TargetFrameworks is not already defined, so something like this:

```
<TargetFrameworks Condition=" '$(TargetFrameworks)' == '' ">net40;net45;netcoreapp2.0</TargetFrameworks>
```

gpetrou:
I added some relative path references for the ClrAssembly build. Not sure if this is the best approach.

PR #87, opened 2017-12-14
Target .NET Standard 1.3 on CustomTasks

jp2masa:
## Changes
- Added a csproj for CustomTasks, which builds a netstandard1.3 assembly.
- Updated Bootstrap.targets to call MSBuild instead of Csc.


slide:
Is there a reason to use netstandard1.3 instead of netstandard2.0?

jp2masa:
There are many reasons, but the main one is that MSBuild doesn't target .NET Core 2.0 yet (it will on 15.6 I think).

slide:
Ah, ok, so this is specifically because it's a custom task for msbuild. Understoond

jp2masa:
It's also a good practice to target the smallest .NET Standard version possible, so that it works on more platforms.

PR #88, opened 2017-12-15
Clear intepreter stack after pop

slozier:
See https://github.com/IronLanguages/ironpython2/issues/273

### DIFF
--- a/Tests/Microsoft.Dynamic.Test/InterpreterTest.cs
+++ b/Tests/Microsoft.Dynamic.Test/InterpreterTest.cs
@@ -1,0 +1,38 @@
+﻿using NUnit.Framework;
+using System;
+using System.Linq.Expressions;
+
+namespace Microsoft.Dynamic.Test {
+    [TestFixture]
+    public class InterpreterTest {
+        public class TestGc {
+            static int cnt = 0;
+
+            public static void CheckCount() {
+                GC.Collect();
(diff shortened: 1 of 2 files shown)


slide:
Is there any sort of test we could add here? I don't have anything specific in mind, it would just be nice to imrove the dlr testing.

slozier:
I have no idea, does the DLR have a builtin language we can use for testing?

slide:
We could add some nunit stuff for testing. We run some tests, but they are just programs that are run.

slozier:
Hmm, interesting the test failed on Travis for netcoreapp. Not sure how I'll go about figuring that one out...

Tests/Microsoft.Dynamic.Test/InterpreterTest.cs
```diff
@@ -0,0 +1,41 @@
+﻿using NUnit.Framework;
+using System;
+using System.Linq.Expressions;
+
+namespace Microsoft.Dynamic.Test {
+    [TestFixture]
+    public class InterpreterTest {
+        [TestFixture]
+        public class MathUtilsTest {
```
gpetrou: Is this needed or a copy/paste error?

slozier: Yes, good catch.

PR #89, opened 2017-12-16
Remove BigInteger class

gpetrou:


Src/Microsoft.Dynamic/Ast/ConstantExpression.cs
```diff
@@ -55,26 +56,39 @@ public static partial class Utils {
                 return NullLiteral;
             }
 
-            BigInteger bi = value as BigInteger;
-            if ((object)bi != null) {
-                return BigIntegerConstant(bi);
-            } else if (value is BigInt) {
-                return BigIntConstant((BigInt)value);
-            } else if (value is Complex) {
+            if (value is BigInteger) {
```
slozier: What about using a switch statement instead? Removing the redundant elses ends up taking something quite compact and makes it take a lot more space.

Src/Microsoft.Dynamic/Ast/ConstantExpression.cs
```diff
@@ -55,26 +56,39 @@ public static partial class Utils {
...
+            }
+
+            if (value is MethodInfo) {
                 return Expression.Constant(value, typeof(MethodInfo));
-            } else if (value is PropertyInfo) {
+            }
+
+            if (value is PropertyInfo) {
                 return Expression.Constant(value, typeof(PropertyInfo));
             } else {
```
slozier: If you're removing redundant elses then this one should go as well.

Src/Microsoft.Dynamic/Ast/ConstantExpression.cs
```diff
@@ -104,49 +118,25 @@ public static partial class Utils {
...
-
-            return Expression.Call(
-                new Func<bool, byte[], BigInteger>(CompilerHelpers.CreateBigInteger).GetMethodInfo(),
-                Constant(value.Sign < 0),
-                CreateArray<byte>(value.Abs().ToByteArray())
-            );
-        }
-
-        private static Expression BigIntConstant(BigInt value) {
+        private static Expression BigIntConstant(BigInteger value) {
```
slozier: Rename to `BigIntegerConstant`?

Src/Microsoft.Dynamic/Generation/CompilerHelpers.cs
```diff
@@ -857,19 +856,19 @@ private sealed class DebuggableCodeRewriter : DynamicExpressionVisitor {
         #region Factories
 
         public static BigInteger CreateBigInteger(bool isNegative, byte[] data) {
```
slozier: This method should be removed.

Src/Microsoft.Dynamic/Generation/CompilerHelpers.cs
```diff
@@ -857,19 +856,19 @@ private sealed class DebuggableCodeRewriter : DynamicExpressionVisitor {
...
-            return new BigInteger(CreateBigInt(isNegative, data));
+            return CreateBigInt(isNegative, data);
         }
 
-        public static BigInt CreateBigInt(int value) {
-            return (BigInt)value;
+        public static BigInteger CreateBigInt(int value) {
```
slozier: I know these are public, but I'm wondering if we should still rename them to `CreateBigInteger`...

gpetrou: Not sure. Do you still want me to do this change?

slide: I would like to mark this one as Obsolete and add `CreateBigInteger` and rename all internal and IronPython usages to use `CreateBigInteger`. We can do as a separate PR if you don't want to do it in this one.

PR #92, opened 2017-12-17
Project file cleanup

jp2masa:
## Changes
- Project file cleanup.

### DIFF
--- a/Build.proj
+++ b/Build.proj
@@ -1,17 +1,17 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003" ToolsVersion="4.0">
+
   <PropertyGroup>
     <BuildFlavour Condition=" '$(BuildFlavour)' == '' ">Debug</BuildFlavour>
     <Mono Condition="'$(OS)' != 'Windows_NT'">true</Mono>
 
     <RootDir>$(MSBuildThisFileDirectory)</RootDir>
     <Solution>$(RootDir)\Dlr.sln</Solution>
-    
+
     <UtilDir>$(RootDir)\Util</UtilDir>
     <BinDir>$(RootDir)\bin</BinDir>
     <BuildSysDir>$(RootDir)\Build</BuildSysDir>
     <MSBuildCommunityTasksPath>$(RootDir)\Util\MSBuildCommunityTasks</MSBuildCommunityTasksPath>
-    <TargetFrameworks Condition=" '$(TargetFrameworks)' == '' ">net40;net45;netcoreapp2.0</TargetFrameworks>
   </PropertyGroup>
 
   <Import Project="$(RootDir)\CurrentVersion.props" />
@@ -25,7 +25,7 @@
     <StageDir>$(RootDir)\Package\$(BuildFlavour)\Stage\DynamicLanguageRuntime.$(PackageVersion)</StageDir>
     <PackageDir>$(RootDir)\Package\$(BuildFlavour)\Packages\DynamicLanguageRuntime.$(PackageVersion)</PackageDir>
   </PropertyGroup>
-  
+
   <PropertyGroup>  
     <BuildProperties>
       RootDir=$(RootDir);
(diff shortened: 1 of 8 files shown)


Build.proj
```diff
@@ -1,17 +1,17 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003" ToolsVersion="4.0">
+
   <PropertyGroup>
     <BuildFlavour Condition=" '$(BuildFlavour)' == '' ">Debug</BuildFlavour>
-    <Mono Condition="'$(OS)' != 'Windows_NT'">true</Mono>
+    <Mono Condition="'$(MSBuildRuntimeType)' == 'Mono'">true</Mono>
```
jp2masa: Not sure about this change. If `$(OS)` is not `Windows_NT`, it doesn't necessarily mean it's running on Mono, but is it possible that `$(MSBuildRuntimeType)` is not `Mono`, when building for Mono?

jp2masa:
Project files used to be edited by UIs, but the new csproj format tends to be edited by hand more frequently. I made that change to improve project file readability, but I can remove the newlines if you prefer.

slide:
Not yet, let me let it summer a bit and see what I think first. Thanks for the improvements.

Src/Microsoft.Dynamic/Microsoft.Dynamic.csproj
```diff
@@ -1,39 +1,46 @@
...
   <Import Project="$(ProjectDir)..\..\Build\Common.proj" />
+
   <PropertyGroup>
-    <TargetFrameworks Condition="'$(TargetFrameworks)' == ''">net40;net45;netcoreapp2.0</TargetFrameworks>
-    <PlatformTarget>anycpu</PlatformTarget>
-    <ProjectGuid>{EB66B766-6354-4208-A3D4-AACBDCB5C3B3}</ProjectGuid>
-    <OutputType>Library</OutputType>
-    <AppDesignerFolder>Properties</AppDesignerFolder>
+    <TargetFrameworks>net40;net45;netcoreapp2.0</TargetFrameworks>
     <RootNamespace>Microsoft.Scripting</RootNamespace>
```
slozier: Not an issue with this PR, but I don't think should be Microsoft.Scripting. We can probably get rid of it.

Build.proj
slozier: Would be nice to not have a compile time Mono check and just do it at runtime.

slide: Agreed, it shouldn't be that difficult to do.

Src/Microsoft.Dynamic/Microsoft.Dynamic.csproj
jp2masa: All the files in this project use `Microsoft.Scripting` as the root namespace. Removing this property means that when creating new files in this project the namespace will be based on `Microsoft.Dynamic` instead.

Build.proj
jp2masa: It looks like the `Mono` property is defined here too: https://github.com/IronLanguages/dlr/blob/master/Build/Common.proj#L10.

The `Mono` property is currently used to define a flag which is used in one place only: https://github.com/IronLanguages/dlr/blob/340682806d1cd54d72147c7f2f8f5a120c1ba6be/Src/Microsoft.Dynamic/Interpreter/InterpretedFrame.cs#L139-L143
You mean checking at runtime like [this](https://stackoverflow.com/a/721194/4647866)?

slozier: Yes, that type runtime check.

Src/Microsoft.Dynamic/Microsoft.Dynamic.csproj
slozier: You're right, I hadn't checked the source files. We can leave it as is.

Build.proj
slide: We have a check in IronPython right now [here](https://github.com/IronLanguages/ironpython2/blob/40f51b92f8294bb901d29738707103f92fe19c87/Src/IronPython/Runtime/ClrModule.cs#L67). Honestly, it would be nice to migrate all the feature checking into one place, but I haven't thought about how I really want to do that yet.

jp2masa: Should I revert the change on this line or move the Mono check to runtime?

slide: Let's revert for now and add something as a runtime check in a different PR

PR #93, opened 2017-12-20
ClrAssembly improvements

jp2masa:
## Changes
- ClrAssembly improvements:
  - Incremental targets.
  - Better support for the dotnet cli.
  - Clean deletes all generated assemblies.
- Fixed Dlr.sln broken paths.

### DIFF
--- a/Tests/ClrAssembly/ClrAssembly.csproj
+++ b/Tests/ClrAssembly/ClrAssembly.csproj
@@ -48,17 +48,13 @@
     <Csc TargetType="Library" References="@(ReferencePath);$(OutputPath)\$(Prefix)typesamples.dll" Sources="%(AreaCSFile.Identity)" OutputAssembly="$(OutputPath)\$(Prefix)%(AreaCSFile.Filename).dll" />
   </Target>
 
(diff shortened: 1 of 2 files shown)


Tests/ClrAssembly/ClrAssembly.csproj
```diff
@@ -4,7 +4,7 @@
 
   <PropertyGroup>
     <TargetFrameworks>net40;net45;netcoreapp2.0</TargetFrameworks>
-    <PlatformTarget Condition="'$(TargetFramework)' == 'netcoreapp2.0'">x64</PlatformTarget>
+    <PlatformTarget Condition="'$(TargetFramework)' == 'netcoreapp2.0' AND '$(OS)' != 'Windows_NT'">x64</PlatformTarget>
```
jp2masa: Not fully sure about this change, but I think that x86 Windows is supported by .NET Core.

slide: The problem is specifying Any CPU, that doesn't work.

jp2masa: I can't understand why it doesn't work. It doesn't work on Mono? AnyCPU is the default in the dotnet sdk and it usually works.

slozier: Do we even need `PlatformTarget` in `ClrAssembly`? Doesn't it only affect the dummy assembly which we don't use?

PR #98, opened 2017-12-29
Replace is operator with switch pattern matching

gpetrou:


Src/Microsoft.Dynamic/Actions/Calls/DefaultOverloadResolver.cs
```diff
@@ -220,16 +220,17 @@ public DefaultOverloadResolver(ActionBinder binder, IList<DynamicMetaObject> arg
             IDictionaryEnumerator dictEnum = dict.GetEnumerator();
             while (dictEnum.MoveNext()) {
                 DictionaryEntry de = dictEnum.Entry;
-                
-                if (de.Key is string) {
-                    splattedNames.Add((string)de.Key);
+
+                string s = de.Key as string;
```
slozier: If we're going to change this I would prefer using pattern matching instead
```C#
if (de.Key is string s) {
    // ...
}
```

slide:
Just want to double check, have all the issues been resolved on this?

gpetrou:
If I remember correctly, I did the changes that @slozier had asked.

PR #111, opened 2018-04-18
Make all libs support netstandard2.0

LonghronShen:
It's not a good idea for a lib only support netcoreapp2.0 profile.
I changed it to netstandard2.0  to let the libs use as less dependencies as possible.

### DIFF
--- a/Src/Microsoft.Dynamic/Microsoft.Dynamic.csproj
+++ b/Src/Microsoft.Dynamic/Microsoft.Dynamic.csproj
@@ -1,46 +1,50 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
-  <Import Project="$(ProjectDir)..\..\Build\Common.proj" />
+    <Import Project="$(ProjectDir)..\..\Build\Common.proj" />
 
-  <PropertyGroup>
-    <TargetFrameworks>net45;netcoreapp2.0</TargetFrameworks>
-    <RootNamespace>Microsoft.Scripting</RootNamespace>
-    <BaseAddress>859832320</BaseAddress>
-    <CodeAnalysisRuleSet>$(AssemblyName).ruleset</CodeAnalysisRuleSet>
-    <DocumentationFile>$(OutputPath)\$(TargetFramework)\$(AssemblyName).xml</DocumentationFile>
-  </PropertyGroup>
+    <PropertyGroup>
+        <TargetFrameworks>net45;netstandard2.0</TargetFrameworks>
+        <RootNamespace>Microsoft.Scripting</RootNamespace>
+        <BaseAddress>859832320</BaseAddress>
+        <CodeAnalysisRuleSet>$(AssemblyName).ruleset</CodeAnalysisRuleSet>
+        <DocumentationFile>$(OutputPath)\$(TargetFramework)\$(AssemblyName).xml</DocumentationFile>
+    </PropertyGroup>
 
-  <ItemGroup>
-    <Compile Include="..\AssemblyInfo.cs" Link="AssemblyInfo.cs" />
-  </ItemGroup>
+    <PropertyGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+        <DefineConstants>$(DefineConstants.Replace(';FEATURE_FILESYSTEM', ''))</DefineConstants>
+    </PropertyGroup>
 
-  <ItemGroup>
-    <ProjectReference Include="..\Microsoft.Scripting.Metadata\Microsoft.Scripting.Metadata.csproj" Condition=" $(Features.Contains('FEATURE_METADATA_READER')) " />
-    <ProjectReference Include="..\Microsoft.Scripting\Microsoft.Scripting.csproj" />
-  </ItemGroup>
+    <ItemGroup>
+        <Compile Include="..\AssemblyInfo.cs" Link="AssemblyInfo.cs" />
+    </ItemGroup>
 
-  <ItemGroup>
-    <Reference Include="System.Configuration" Condition=" $(Features.Contains('FEATURE_CONFIGURATION')) " />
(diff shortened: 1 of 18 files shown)


slide:
There is a specific reason we don't don't target netstandard2.0. It doesn't include everything needed at this point.

LonghronShen:
In theory, the netcoreapp2.0 profile is just a big bundle of packages.
If you wish, you can add any required package one by one in addition to netstandard2.0 profile. The only magic thing about netcoreapp2.0 is that it allows the project to be launched as a standalone app.
I just think the netcoreapp2.0 contains too many packages that we don't need if you are creating a new lib referencing this lib. Another, not all the final entrypoint projects are console applications, e.g., WPF or Xamarin apps. These types of projects don't consume netcoreapp2.0 profile.

slide:
That's not completely true. There are reflection apis that don't exist in netstandard2.0 available libs.

LonghronShen:
My fork works well on netstandard2.0 profile now...

slide:
By the way, for WPF you would use the next targeted libraries that are available in nuget as well

LonghronShen:
Good point, I missed the net45 profile support, lol.

slide:
Regardless though, this change caused test failures so it can't be merged.

LonghronShen:
OK  all tests cleared!

Src/Microsoft.Dynamic/Utils/ReflectionUtils.cs
```diff
@@ -263,7 +263,7 @@ private static IEnumerable<T> WithBindingFlags<T>(this IEnumerable<T> members, F
         //
         // The general algorithm that is used for determining the names in a type and the layout of objects of the type is roughly as follows:
         // - Flatten the inherited names (using the hide by name or hide by name-and-signature rule) ignoring accessibility rules. 
-        // - For each new member that is marked �expect existing slot�, look to see if an exact match on kind (i.e., field or method),
```
slide: This should not be changed.

Src/Microsoft.Dynamic/Utils/ReflectionUtils.cs
```diff
@@ -325,7 +325,7 @@ private static IEnumerable<T> WithBindingFlags<T>(this IEnumerable<T> members, F
         // -------------------------
         // [Note: The CLS (see Partition I) refers to instance, virtual, and static properties.  
         // The signature of a property (from the Type column) can be used to distinguish a static property, 
-        // since instance and virtual properties will have the �HASTHIS� bit set in the signature (�23.2.1)
+        // since instance and virtual properties will have the �HASTHIS?bit set in the signature (?3.2.1)
```
slide: This should not be changed.

Src/Microsoft.Scripting/Microsoft.Scripting.csproj
```diff
@@ -1,39 +1,40 @@
...
-    <Compile Include="..\AssemblyInfo.cs" Link="AssemblyInfo.cs" />
-  </ItemGroup>
+    <ItemGroup>
+        <Compile Include="..\AssemblyInfo.cs" Link="AssemblyInfo.cs" />
+    </ItemGroup>
 
-  <ItemGroup>
-    <Reference Include="System.Configuration" Condition=" $(Features.Contains('FEATURE_CONFIGURATION')) " />
-  </ItemGroup>
+    <ItemGroup Condition="'$(TargetFramework)' == 'net45'">
```
slide: This shouldn't be needed if the props file is correct.

Src/Microsoft.Dynamic/Utils/ReflectionUtils.cs
LonghronShen: Maybe my VS2017 code format destroyed some special characters...
Let me try to restore it...

LonghronShen: The problem comes from my Git Bash for Windows. It clones the repo and converts the encoding of the file from UTF8 to ASCII based local code page.

Src/Microsoft.Scripting/Microsoft.Scripting.csproj
slide: I disagree, please keep the method we were originally using.

slozier:
I hate to say this over and over again, but even though System.Reflection.Emit package says it works with netstandard1.1, it is not a netstandard package and only works with .NET Core. I am not particularly interested in saying that we're running on netstandard2.0 when in fact the netstandard2.0 package only works with netcoreapp2.0...

Src/Microsoft.Dynamic/Microsoft.Dynamic.csproj
```diff
@@ -1,46 +1,50 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
-  <Import Project="$(ProjectDir)..\..\Build\Common.proj" />
+    <Import Project="$(ProjectDir)..\..\Build\Common.proj" />
 
-  <PropertyGroup>
-    <TargetFrameworks>net45;netcoreapp2.0</TargetFrameworks>
```
slozier: The netcoreapp2.0 targets should not be removed.

Src/Microsoft.Scripting/PlatformAdaptationLayer.cs
```diff
@@ -54,7 +54,7 @@ public class PlatformAdaptationLayer {
 
         #region Assembly Loading
 
-#if NETCOREAPP2_0
+#if NETCOREAPP2_0 || NETSTANDARD2_0
```
slozier: This is a .NET Core specific thing.

slozier:
The `System.Reflection.Emit` package has been unlisted so until MS comes up with something netstandard2.0 is not an option.

PR #113, opened 2018-05-13
Dispose IEnumerator instances

gpetrou:


Src/Microsoft.Dynamic/Debugging/CollectionUtils.cs
```diff
@@ -30,16 +30,24 @@ internal static class CollectionUtils {
...
             var cmp = EqualityComparer<T>.Default;
-            var f = first.GetEnumerator();
-            var s = second.GetEnumerator();
-            while (f.MoveNext()) {
-                s.MoveNext();
 
-                if (!cmp.Equals(f.Current, s.Current)) {
-                    return false;
+            IEnumerator<T> f = null;
+            try {
```
slozier: Why use a try/finally for the first enumerator instead of two usings?

gpetrou: I had https://msdn.microsoft.com/en-us/library/ms182334.aspx in mind but looking at this again we can use two usings since they are independent of each other.

Src/Microsoft.Dynamic/Debugging/CollectionUtils.cs
```diff
@@ -29,17 +29,22 @@ internal static class CollectionUtils {
...
+
             var cmp = EqualityComparer<T>.Default;
-            var f = first.GetEnumerator();
-            var s = second.GetEnumerator();
-            while (f.MoveNext()) {
-                s.MoveNext();
 
-                if (!cmp.Equals(f.Current, s.Current)) {
-                    return false;
+            using (var f = first.GetEnumerator()) {
```
slozier: You can get rid of the extra indent/brackets:
```C#
using (var f = first.GetEnumerator())
using (var s = second.GetEnumerator()) {
        // ...
}
```

gpetrou: I would prefer to keep the brackets if you don't mind.

PR #114, opened 2018-05-13
Use nameof

gpetrou:


Src/Microsoft.Dynamic/Utils/ContractUtils.cs
```diff
@@ -49,9 +49,13 @@ public static class ContractUtils {
             }
         }
 
+        public static void RequiresNotNull(object value) {
+            string paramName = nameof(value);
```
slozier: I don't think this is correct. Doesn't `nameof(value) == "value"`?

gpetrou: My bad. I will update.

PR #121, opened 2018-06-11
Use pattern matching and inline out parameters

gpetrou:


Src/Microsoft.Scripting/Hosting/ExceptionOperations.cs
```diff
@@ -70,8 +70,7 @@ public sealed class ExceptionOperations : MarshalByRefObject {
 
         public bool HandleException(ObjectHandle exception) {
             ContractUtils.RequiresNotNull(exception, nameof(exception));
-            var exceptionObj = exception.Unwrap() as Exception;
-            ContractUtils.Requires(exceptionObj != null, "exception", "ObjectHandle must be to Exception object");
+            ContractUtils.Requires(exception.Unwrap() is Exception exceptionObj, "exception", "ObjectHandle must be to Exception object");
```
slide: nameof(exception) perhaps?

slozier: also no point in assigning to `exceptionObj` if it's unused?

Src/Microsoft.Scripting/Hosting/ExceptionOperations.cs
```diff
@@ -55,30 +55,30 @@ public sealed class ExceptionOperations : MarshalByRefObject {
...
-            ContractUtils.Requires(exceptionObj != null, "exception", "ObjectHandle must be to Exception object");
+            ContractUtils.Requires(exceptionObj != null, nameof(exception), "ObjectHandle must be to Exception object");
 
             _context.GetExceptionMessage(exceptionObj, out message, out errorTypeName);
         }
 
         public bool HandleException(ObjectHandle exception) {
             ContractUtils.RequiresNotNull(exception, nameof(exception));
-            ContractUtils.Requires(exception.Unwrap() is Exception exceptionObj, "exception", "ObjectHandle must be to Exception object");
+            ContractUtils.Requires(exception.Unwrap() is Exception _, nameof(exception), "ObjectHandle must be to Exception object");
```
slide: You don't need the _ at all if it just wants a boolean expression.

PR #124, opened 2018-06-14
Add netcoreapp2.1

slozier:


Tests/ClrAssembly/ClrAssembly.csproj
```diff
@@ -3,8 +3,7 @@
   <Import Project="$(ProjectDir)..\..\Build\Common.proj" />
 
   <PropertyGroup>
-    <TargetFrameworks>net45;netcoreapp2.0</TargetFrameworks>
-    <PlatformTarget Condition="'$(TargetFramework)' == 'netcoreapp2.0' AND '$(OS)' != 'Windows_NT'">x64</PlatformTarget>
```
slide: Did you test with IronPython with this change? I think this was added because IronPython needs the x64 build for some reason.

slozier: Looks like it's only needed since some assemblies are referenced directly for the unix build? Will investigate.

slozier: I was looking at an old file, the assemblies are no longer referenced directly so it's probably not an issue.

slide:
The issue was specifically running tests on Linux if I remember correctly. `dotnet test` was failing I believe, but maybe its not an issue anymore.

slozier:
Ran the tests locally on Linux, seems to be working fine without the PlatformTarget.

PR #127, opened 2018-06-17
Modernize Microsoft.Scripting/Runtime files

gpetrou:


Src/Microsoft.Scripting/Runtime/ScriptDomainManager.cs
```diff
@@ -42,17 +42,11 @@ public sealed class ScriptDomainManager {
...
 
-        public SharedIO SharedIO {
-            get { return _sharedIO; }
-        }
+        public SharedIO SharedIO { get; }
 
-        public DynamicRuntimeHostingProvider Host {
-            get { return _hostingProvider; }
-        }
+        public DynamicRuntimeHostingProvider Host => _hostingProvider;
```
slozier: Use auto property instead?

Src/Microsoft.Scripting/Runtime/ScriptDomainManager.cs
```diff
@@ -42,17 +42,11 @@ public sealed class ScriptDomainManager {
...
 
-        public DlrConfiguration Configuration {
-            get { return _configuration; }
-        }
+        public DlrConfiguration Configuration => _configuration;
```
slozier: auto property?

Src/Microsoft.Scripting/Runtime/TokenInfo.cs
```diff
@@ -25,18 +25,18 @@ public struct TokenInfo : IEquatable<TokenInfo> {
         private SourceSpan _span;
         
         public TokenCategory Category {
```
slozier: Why not make all of these auto properties?

Src/Microsoft.Scripting/Runtime/ContextId.cs
```diff
@@ -57,21 +55,16 @@ public struct ContextId : IEquatable<ContextId> {
...
-            return ContextId.Empty;
+            return Empty;
         }
 
-        public int Id {
-            get {
-                return _id;
-            }
-        }
+        public int Id => _id;
```
slozier: Make it an auto property?

Src/Microsoft.Scripting/Runtime/DlrConfiguration.cs
```diff
@@ -30,18 +31,12 @@ internal sealed class LanguageConfiguration {
...
         
-        public LanguageContext LanguageContext {
-            get { return _context; }
-        }
+        public LanguageContext LanguageContext => _context;
 
-        public AssemblyQualifiedTypeName ProviderName {
-            get { return _providerName; }
-        }
+        public AssemblyQualifiedTypeName ProviderName => _providerName;
```
slozier: These could be auto properties...

Src/Microsoft.Scripting/Runtime/NotNullAttribute.cs
```diff
@@ -32,7 +32,5 @@ public sealed class NotNullAttribute : Attribute {
     /// </summary>
     [AttributeUsage(AttributeTargets.Parameter, AllowMultiple = false, Inherited = false)]
     public sealed class NotNullItemsAttribute : Attribute {
-        public NotNullItemsAttribute() {
```
slozier: Although unlikely, if you remove the this constructor and someone comes along and add a constructor overload (e.g. with an argument) then the default constructor will be gone.

Src/Microsoft.Scripting/Runtime/ScriptCode.cs
```diff
@@ -31,13 +31,9 @@ public abstract class ScriptCode {
...
 
-        public LanguageContext LanguageContext {
-            get { return _sourceUnit.LanguageContext; }
-        }
+        public LanguageContext LanguageContext => _sourceUnit.LanguageContext;
 
-        public SourceUnit SourceUnit {
-            get { return _sourceUnit; }
-        }
+        public SourceUnit SourceUnit => _sourceUnit;
```
slozier: auto property?

Src/Microsoft.Scripting/Runtime/NotNullAttribute.cs
gpetrou: I think we had removed the NotNullAttribute constructor in the same file sometime ago as well, so this makes the changes consistent.

PR #147, opened 2018-07-22
Convert AddInstruction classes closer to .NET Core implementation

gpetrou:
I replaced the existing class code with the corresponding .NET Core code minus the following:

- They use different variable names, s_Int16 instead of _Int16 for example.
- They use extension methods such as GetNonNullableType and IsArithmetic. Maybe we want to change them in separate PR for all files?

Also, the order of some methods is different. Do we want to keep our order?

Finally, they use InstructionName instead of overriding ToString. Do we want that change?

### DIFF
--- a/Src/Microsoft.Dynamic/ComInterop/VarEnumSelector.cs
+++ b/Src/Microsoft.Dynamic/ComInterop/VarEnumSelector.cs
@@ -341,7 +341,7 @@ namespace Microsoft.Scripting.ComInterop {
 
             // COM cannot express valuetype nulls so we will convert to underlying type
             // it will throw if there is no value
-            if (TypeUtils.IsNullableType(argumentType)) {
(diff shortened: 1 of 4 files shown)


slozier:
My preference would be to keep the two in sync as much as possible (static variable names, order of methods, etc). That way we can more easily bring in changes (or submit improvements) to .NET Core.

I believe we can change GetNonNullableType and IsArithmetic to an extension method without having to change any of the other source files. We should check that the implementations of these methods are in sync with the ones used by .NET Core.

I looked over the code base before and I believe the same is true for all the arithmetic instructions.

The default `ToString` is just `InstructionName + "()"` so we can probably use whatever they have.

gpetrou:
I only applied the method changes for now.
I will do some other items, such as the InstructionName changes in a separate PR.

PR #149, opened 2018-07-28
[WIP] Fix TestResults

gpetrou:


slide:
We use dotnet test and generate trx files, so I am not sure why this is needed.

slide:
Also, you need to always use / because the steps are shared between Windows and Unix

slide:
Can we close this out?

gpetrou:
Oh, I see.
I thought we were using the NUnit runner to run the tests.

PR #151, opened 2018-07-29
Convert LessThanInstruction classes closer to .NET Core implementation

gpetrou:


Src/Microsoft.Dynamic/Interpreter/Instructions/LessThanInstruction.cs
```diff
@@ -8,118 +8,238 @@
 using Microsoft.Scripting.Utils;
 
 namespace Microsoft.Scripting.Interpreter {
-    public abstract class LessThanInstruction : Instruction {
+    internal abstract class LessThanInstruction : Instruction {
+        public static readonly object BoxedFalse = false;
+        public static readonly object BoxedTrue = true;
```
gpetrou: Where should we place the above two lines? In a new class somewhere?

slozier: There are ScriptingRuntimeHelpers.True/False which we should use instead.

slide:
Do they have additional tests for these things as well?

slide:
I found this: https://github.com/dotnet/corefx/blob/master/src/System.Linq.Expressions/tests/InterpreterTests.cs

gpetrou:
Looking through .NET Core repository, I don't see dedicated tests to Instruction classes, such as the one in this PR.
Should we add the Interpreter test in a separate PR?

slozier:
I don't think we can use the InterpreterTests from .NET Core as is. Better to leave it for another PR. We should look into creating some basic tests for the instructions though.

Src/Microsoft.Dynamic/Interpreter/Instructions/LessThanInstruction.cs
slide: @gpetrou Are you still wanting this merged? Can you update to use the ScriptingRuntimeHelpers?

gpetrou: @slide I did as you asked. Do you want to merge this?

slozier:
@gpetrou Thanks for the PR, sorry it took so long to get merged in.

PR #164, opened 2019-01-02
Remove CurrentVersion and BuildInfo templates

slozier:


Directory.Build.props
```diff
@@ -28,8 +42,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <BeforeTargetFiles>$(DlrBuildDir)\Versioning.targets</BeforeTargetFiles>                                                                                                                                                                 
-    <BeforeTargets>GenerateBuildInfo</BeforeTargets>
+    <BeforeTargets></BeforeTargets>
```
slide: Do we need this here at all if it is empty?

Tests/ClrAssembly/ClrAssembly.csproj
```diff
@@ -24,7 +24,7 @@
     <Csc TargetType="Library" References="@(ReferencePath)" Sources="%(LoadOrderCSFile.Identity)" OutputAssembly="$(OutputPath)\$(Prefix)%(LoadOrderCSFile.Filename).dll" />
   </Target>
 
-  <Target Name="ReachType" Inputs="@(ReachTypeInput)" Outputs="@(ReachTypeOutput)" DependsOnTargets="GetFrameworkPath">
+  <Target Name="ReachType" Inputs="@(ReachTypeInput)" Outputs="@(ReachTypeOutput)" DependsOnTargets="GetFrameworkPath" Condition="'$(MSBuildRuntimeType)' != 'Core'">
```
slide: Is this actually part of this change for removing those files?

slide:
Is `AssemblyConfiguration` automatically generated?

Tests/ClrAssembly/ClrAssembly.csproj
slozier: It's the only change needed to get it building with `dotnet build Dlr.sln` so I threw it in there.

Directory.Build.props
slozier: It's not needed, but I left it in there since we're "using" it with
```
<Target Name="BeforeBuildStarts" BeforeTargets="BeforeBuild" DependsOnTargets="$(BeforeTargets)" />
```

We could probably get rid of both since they do nothing...

slozier:
Yes, it's automatically generated.

Tests/ClrAssembly/ClrAssembly.csproj
slide: :+1:

Directory.Build.props
slide: Yeah, let's remove both if we aren't using them.

PR #176, opened 2019-02-20
Support .Net Standard

YairHalberstadt:
I have tested deploying this as a Nuget Package, referencing and using the nuget package from a .Net Standard class library, and then calling the library from both .Net Core and .Net Framework. Everything works as expected.

I hope that once this is merged, supporting .Net Standard for IronPython should be trivial.

### DIFF
--- a/Package/nuget/DynamicLanguageRuntime.nuspec
+++ b/Package/nuget/DynamicLanguageRuntime.nuspec
@@ -27,8 +27,8 @@
       <group targetFramework="netstandard">
         <dependency id="System.CodeDom" version="4.5.0" />
         <dependency id="System.Configuration.ConfigurationManager" version="4.5.0" />
-        <dependency id="System.Reflection.Emit" Version="4.6.0-preview.19073.11" />
-        <dependency id="System.Reflection.Emit.Lightweight" Version="4.6.0-preview.19073.11" />
+        <dependency id="System.Reflection.Emit" version="4.6.0-preview.19073.11" />
+        <dependency id="System.Reflection.Emit.Lightweight" version="4.6.0-preview.19073.11" />
       </group>
     </dependencies>
   </metadata>
--- a/Package/nuget/DynamicLanguageRuntime.nuspec
+++ b/Package/nuget/DynamicLanguageRuntime.nuspec
@@ -24,7 +24,7 @@
         <dependency id="System.CodeDom" version="4.5.0" />
         <dependency id="System.Configuration.ConfigurationManager" version="4.5.0" />
       </group>
-      <group targetFramework="netstandard">
+      <group targetFramework="netstandard2.0">
         <dependency id="System.CodeDom" version="4.5.0" />
         <dependency id="System.Configuration.ConfigurationManager" version="4.5.0" />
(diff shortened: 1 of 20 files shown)


Package/nuget/DynamicLanguageRuntime.nuspec
```diff
@@ -24,6 +24,12 @@
         <dependency id="System.CodeDom" version="4.5.0" />
         <dependency id="System.Configuration.ConfigurationManager" version="4.5.0" />
       </group>
+      <group targetFramework="netstandard">
+        <dependency id="System.CodeDom" version="4.5.0" />
+        <dependency id="System.Configuration.ConfigurationManager" version="4.5.0" />
+        <dependency id="System.Reflection.Emit" version="4.6.0-preview.19073.11" />
```
YairHalberstadt: I am using the prerelease versions as I believe (but am not sure) that the prerelease versions will throw a PlatformNotSupportedException on AOT platforms, whereas the stable versions wont. I hope to give that a test at some point.

YairHalberstadt: Tested, and I am correct. Stable will throw this exception:

> System.Exception: 'Method 'TypeBuilder.DefineConstructor(MethodAttributes, CallingConventions, Type[])' was not included in compilation, but was referenced in DelegateUtils.EmitCallSiteDelegateType(int). There may have been a missing assembly.'

The preview will throw this, more helpful exception:

> System.PlatformNotSupportedException: PlatformNotSupported_RefEmit. For more information, visit http://go.microsoft.com/fwlink/?LinkId=623485

make.ps1
```diff
@@ -4,7 +4,7 @@ Param(
     [Parameter(Position=1)]
     [String] $target = "release",
     [String] $configuration = "Release",
-    [String[]] $frameworks=@('net45','netcoreapp2.0','netcoreapp2.1'),
+    [String[]] $frameworks=@('net45','netcoreapp2.0','netcoreapp2.1', 'netstandard2.0'),
```
slozier: This list is only used for testing and they're not being build against `netstandard2.0` in this PR so we shouldn't include this.

Src/Microsoft.Dynamic/Generation/ILGen.cs
```diff
@@ -251,7 +251,7 @@ public class ILGen {
             _ilg.EmitCall(opcode, methodInfo, optionalParameterTypes);
         }
 
-#if !NETCOREAPP2_0 || NETCOREAPP2_1
+#if (!NETCOREAPP2_0 && !NETSTANDARD2_0) || NETCOREAPP2_1
```
slozier: Nit: I'm not sure why we had the `|| NETCOREAPP2_1` here. Doesn't do anything and can be removed.

make.ps1
YairHalberstadt: Done

Src/Microsoft.Dynamic/Generation/ILGen.cs
YairHalberstadt: Done

slozier:
Thanks again!

PR #179, opened 2019-03-18
Initial UWP support

DjArt:
What was done:
1. Main projects was additionally targeted to UWP
2. All incompatible `Type` methods and extension was replaced with `TypeInfo` methods and extensions.
`Type` in UWP isn't `MemberInfo`, `TypeInfo` isn't `Type`.
3. Removed part of code, already existed in `System.Reflection`
4. `ReadOnlyCollectionBuilder` was replaced by `List`
5. `ThreadLocal<T>` from DLR was commented for using from `System.Threading`. Need additional testing.
6. `ILGen` was replaced by `ILGenerator`(it's not bad?)
7. Ported tests to UWP and they all is pass.


slozier:
There are a few problems with this PR:
- I don't think targeting `uap10.0` is a good idea, this is essentially the same as trying to target `netstandard1.4` (which we don't want to do). Unfortunately I don't think we can target `uap10.0.16299` (which is `netstandard2.0` compatible) with the current tooling. However I don't see why we can't just use the `netstandard2.0` assemblies?
- Public APIs should not be removed (for example in `ReflectionUtils`)
- There was actually quite a bit of work done to remove `TypeInfo` from the codebase so I'm not inclined to want to add it back in at this point.
- Does not build under CI.
- Tests aren't passing in Release mode?

Anyway, I'd love having things work with UWP but I'm not sure this is the right approach...

DjArt:
1. Test are passing in Release mode, so works after compiling to native. And this is important for the next item.
2. `uap10.0` have differences with any `.netstandard`, they isn't same target. And separating targets is important part of supporting `uap10.0`. Without compilation to native(that is not equal to Release mode), all code runs on `.netcore2.2` and all stuff, even `IronPython` just works. But `uap10.0` limited to native compiler subset. A few points for example:
- `MethodBase` contains `Module` property in `.netstandard2.0`. Native code don't know nothing about `Module`
- `MemberInfo` don't contain `MemberType` or `MetadataToken` in native mode.
If you continue to support the platform without allocating a separate target, it will be extremely difficult for developers.
3. `TypeInfo` at now is one way to escape different code on difference targets.
4. CI not setup for building UWP target on Windows(fails, because tried build package in Release mode for `AnyCPU`) and for ignoring this target on other platforms.
Of course, PR isn't ideal, but it's first steps to support UWP.
@slozier select right way and I will continue help.

DjArt:
@slozier I try to rework PR. It seems possible to do without a `uap10.0` target, however, requires several checks during runtime.
For example, we can place these code somewhere:
```
        public static bool IsNativeModule { get; } = _IsNativeModule();

        private static bool _IsNativeModule() {
            return typeof(void).Assembly.GetType().FullName != "System.Reflection.RuntimeAssembly";
        }
```
And do checks in sensitive code:
Before:
```
public static bool MemberEquals(this MemberInfo self, MemberInfo other) {
            if ((self == null) != (other == null)) {
                // one null, the other isn't.
                return false;
            }

            if (self == null) {
                // both null
                return true;
            }
            if (self.MemberType != other.MemberType) {
                return false;
            }
            switch (self.MemberType) {
                case MemberTypes.Field:
                    return ((FieldInfo)self).FieldHandle.Equals(((FieldInfo)other).FieldHandle);
                case MemberTypes.Method:
                    return ((MethodInfo)self).MethodHandle.Equals(((MethodInfo)other).MethodHandle);
                case MemberTypes.Constructor:
                    return ((ConstructorInfo)self).MethodHandle.Equals(((ConstructorInfo)other).MethodHandle);
                case MemberTypes.NestedType:
                case MemberTypes.TypeInfo:
                    return ((Type)self).TypeHandle.Equals(((Type)other).TypeHandle);
                case MemberTypes.Event:
                case MemberTypes.Property:
                default:
                    return
                        ((MemberInfo)self).Module == ((MemberInfo)other).Module &&
                        ((MemberInfo)self).MetadataToken == ((MemberInfo)other).MetadataToken;
            }
        }
```
After:
```
public static bool MemberEquals(this MemberInfo self, MemberInfo other) {
            if ((self == null) != (other == null)) {
                // one null, the other isn't.
                return false;
            }

            if (self == null) {
                // both null
                return true;
            }

            if (IsNativeModule)
            {
                return self == other;
            }

            if (self.MemberType != other.MemberType) {
                return false;
            }
            switch (self.MemberType) {
                case MemberTypes.Field:
                    return ((FieldInfo)self).FieldHandle.Equals(((FieldInfo)other).FieldHandle);
                case MemberTypes.Method:
                    return ((MethodInfo)self).MethodHandle.Equals(((MethodInfo)other).MethodHandle);
                case MemberTypes.Constructor:
                    return ((ConstructorInfo)self).MethodHandle.Equals(((ConstructorInfo)other).MethodHandle);
                case MemberTypes.NestedType:
                case MemberTypes.TypeInfo:
                    return ((Type)self).TypeHandle.Equals(((Type)other).TypeHandle);
                case MemberTypes.Event:
                case MemberTypes.Property:
                default:
                    return
                        ((MemberInfo)self).Module == ((MemberInfo)other).Module &&
                        ((MemberInfo)self).MetadataToken == ((MemberInfo)other).MetadataToken;
            }
        }
```

slozier:
Targeting netstandard2.0 with a few runtime checks would be the best solution in my opinion. However if we need to have checks all over the place then I would still prefer a separate target.

Just wondering, in your `MemberEquals` example, what is the problematic code?

DjArt:
@slozier problem is get `MemberInfo.MetadataToken`, because https://github.com/dotnet/corert/blob/db9c9ced21afd0848c748960358832bc3e41ae62/src/System.Private.Reflection.Core/src/System/Reflection/Runtime/PropertyInfos/NativeFormat/NativeFormatRuntimePropertyInfo.cs#L126. Yep, I missed with overlapping the right part of the code.

slozier:
@DjArt I see, are these problematic APIs documented or you're just finding them by trial and error?

DjArt:
@slozier unfortunately, they are not documented.

DjArt:
@slozier I'm done. However, this is most likely not all. Further debugging with `IronPython` is required. But, I think, that PR can be closed.

Dlr.sln
```diff
@@ -54,35 +54,49 @@ EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
+		FxCop|Any CPU = FxCop|Any CPU
```
slozier: Could you revert the changes here? We don't use an FxCop configuration anymore.

Src/Microsoft.Scripting/PlatformAdaptationLayer.cs
```diff
@@ -41,6 +41,12 @@ public class PlatformAdaptationLayer {
         [Obsolete("This will be removed in the the future.")]
         public static readonly bool IsCompactFramework = false;
 
+        public static bool IsNativeModule { get; } = _IsNativeModule();
+
+        private static bool _IsNativeModule() {
+            return typeof(void).Assembly.GetType().FullName != "System.Reflection.RuntimeAssembly";
```
slozier: Is there another way to check this? This give a false positive on Mono where `FullName` is `System.Reflection.MonoAssembly`...

DjArt: I'm not sure, but it can be replaced by
`typeof(void).Assembly.Modules.ElementAt(0).ToString() == "<Unknown>"`
https://github.com/dotnet/corert/blob/07fb8bdcae9cabb8f699bc89216dc58909c90e6f/src/System.Private.Reflection.Core/src/System/Reflection/Runtime/Modules/RuntimeModule.cs#L72
There is no more elegant approach, it is necessary to determine on indirect grounds.

slozier:
@DjArt Thanks for the PR!

DjArt:
@slozier I'm glad to help:)

PR #194, opened 2019-05-01
Add .NET Core 3.0 target and enable COM support on that target

jkoritzinsky:
- Add a .NET Core 3.0 target.
- Enable COM on the .NET Core 3.0 target
- Port over the .NET Core COM events implementation
- Make `Microsoft.Scripting.ComInterop.ComObject` public so the `Microsoft.Scripting.ComInterop.ComObject.ObjectToComObject` method is publicly exposed.

Since .NET Core 3.0 will not support using `dynamic` with COM objects because of time constraints (see dotnet/coreclr#24246) and the DLR in this repository already has the a full dynamic COM implementation, I feel that by exposing the `Microsoft.Scripting.ComInterop.ComObject.ObjectToComObject` method so users can hook into this `dynamic` support when using C#'s `dynamic` feature, we can enable users to use `dynamic` with COM objects on .NET Core 3.0.

TODO: Fix Azure DevOps CI to use a host that has VS2019/.NET Core 3.0


Src/Microsoft.Dynamic/ComInterop/ComRuntimeHelpers.cs
```diff
@@ -659,14 +667,13 @@ out argErr
                     typeof(IntPtr), // excepInfo
                     typeof(IntPtr), // argErr
                 };
-            signature.AddArguments(invokeParamTypes, null, null);
-            method.Emit(OpCodes.Calli, signature);
+            method.EmitCalli(OpCodes.Calli, CallingConvention.StdCall, typeof(int), invokeParamTypes);
```
slozier: Any reason for the change of calling convention here?

jkoritzinsky: I was debugging something at the time. I'll change it back to `WinApi`.

slozier:
CI has been having issues lately, triggered a build manually: https://dev.azure.com/dotnet/IronLanguages/_build/results?buildId=16439

This will likely break the IronPython build process since there are some things that are skipped when building with MSBuild Core.

jkoritzinsky:
I'm fine switching back to full MSBuild. I was just having a few issues so I decided to try out the Core version.

jkoritzinsky:
Yeah I don't know why the non-Windows build is failing the way it is. Looks like a bug in Mono (and for somre reason it also looks like the Mono version in those images got downgraded for some reason)

slozier:
I had similar issues with Mono and recent SDKs: https://github.com/dotnet/core/issues/2540

slozier:
If I recall correctly the blocker for switching to building with MSBuild Core is the exec ilasm in ClrAssembly.csproj.

jkoritzinsky:
@slozier where is the `ReachType` target called in the build process?

jkoritzinsky:
Idea: depending on how you do your NuGet deployment, we could change the build to use Full MSBuild on Windows and MSBuild Core off-Windows.

slozier:
@jkoritzinsky It's not used in the build process itself. The assembly is only used for testing.

jkoritzinsky:
@slozier I have a clean CI build using Full MSBuild on Windows and Core MSBuild off-Windows. The bug with using Full MSBuild + .NET Core 3.0 off-Windows that you linked should be fixed in the next VS for Mac release based on what I can find online (it's a versioning bug in NuGet's assemblies).

slozier:
@jkoritzinsky Thanks for the PR. I will have to run some tests against IronPython to make sure this doesn't break anything unexpected (may take a few days before I egt around to it since I'll be at Pycon).

jkoritzinsky:
That’s fine with me. Enjoy Pycon!

PR #206, opened 2019-07-13
Use in-modifier with readonly structs

gpetrou:


slozier:
Unfortunately a lot of these are binary breaking change (see https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/version-update-considerations) which makes it really hard to release the DLR independently of IronPython. I'd like to try and be more careful about this type of thing going forward. Maybe I've just been watching too many .NET Core API reviews...

That being said, I have no problem adding `in` modifiers to anything that is private/internal.

slozier:
Thanks! Looks good!

PR #218, opened 2019-10-13
Replace Hashtables with generic Dictionaries

gpetrou:


Src/Microsoft.Dynamic/ComInterop/ComTypeDesc.cs
```diff
@@ -5,24 +5,18 @@
...
+using System.Globalization;
 using System.Runtime.InteropServices.ComTypes;
 using System.Threading;
 
 namespace Microsoft.Scripting.ComInterop {
 
     public class ComTypeDesc : ComTypeLibMemberDesc {
-        private string _typeName;
-        private string _documentation;
-        //Hashtable is threadsafe for multiple readers single writer.
```
slozier: I have not looked at usage. But I assume this comment is here because it's intended to be used in multithreaded scenarios? Dictionary doesn't provide the same thread safety guarantees as Hashtable.

gpetrou: I used ConcurrentDictionaries instead. What do you think?

slozier: Using ConcurrentDictionary seems fine from a thread safety point of view. Hopefully it isn't a perf hit. Just need to find time to look at how these classes are used and will merge it in assuming there are no issues.

Lamparter:
This should probably be closed 🙂

BCSharp:
I intended to do some performance measurements first to reach the right decision, merge or not.

slozier:
Thanks! Sorry it took so long. 🥲

PR #240, opened 2020-10-26
Support ParamDictionaryAttribute on IReadOnlyDictionary

BCSharp:
This, when completed, will facilitate IronLanguages/ironpython3#882.

I implemented support only for `IReadOnlyDictionary<string, anything>` and not for `IReadOnlyDictionary<object, anything>` like for other generic types supported, because I fail to see any benefit of supporting dictionary interfaces that take `object` type as key, and sticking to `string` eliminates one cast for every runtime call for every kwarg param. The whole param dictionary machinery in DLR assumes that keyword parameter names are strings. If the client code really needs `<object, anything>` specialization to e.g. pass it on somehere down the road, it can always simply declare `Dictionary<object, something>`.

### DIFF
--- a/Src/Microsoft.Dynamic/Utils/CollectionExtensions.cs
+++ b/Src/Microsoft.Dynamic/Utils/CollectionExtensions.cs
@@ -141,11 +141,14 @@ namespace Microsoft.Scripting.Utils {
         internal static ReadOnlyCollection<T> Instance = new ReadOnlyCollection<T>(new T[0]);
     }
 
-#if NETCOREAPP
     internal static class EmptyReadOnlyDictionary<TKey, TValue> {
-        internal static IReadOnlyDictionary<TKey, TValue> Instance = System.Collections.Immutable.ImmutableDictionary.Create<TKey, TValue>();
-    }
+        internal static readonly IReadOnlyDictionary<TKey, TValue> Instance
+#if NETCOREAPP
+            = System.Collections.Immutable.ImmutableDictionary.Create<TKey, TValue>();
+#else
+            = new ReadOnlyDictionary<TKey, TValue>(new Dictionary<TKey, TValue>());
 #endif
+    }
 
     // TODO: Should we use this everywhere for empty arrays?
     // my thought is, probably more hassle than its worth
--- a/Src/Microsoft.Dynamic/Utils/CollectionExtensions.cs
+++ b/Src/Microsoft.Dynamic/Utils/CollectionExtensions.cs
@@ -140,6 +140,13 @@ namespace Microsoft.Scripting.Utils {
     internal static class EmptyReadOnlyCollection<T> {
         internal static ReadOnlyCollection<T> Instance = new ReadOnlyCollection<T>(new T[0]);
     }
+
+#if NETCOREAPP
+    internal static class EmptyReadOnlyDictionary<TKey, TValue> {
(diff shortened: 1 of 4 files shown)


Src/Microsoft.Dynamic/Utils/CollectionExtensions.cs
```diff
@@ -140,6 +140,13 @@ internal static class CollectionExtensions {
     internal static class EmptyReadOnlyCollection<T> {
         internal static ReadOnlyCollection<T> Instance = new ReadOnlyCollection<T>(new T[0]);
     }
+
+#if NETCOREAPP
+    internal static class EmptyReadOnlyDictionary<TKey, TValue> {
+        internal static IReadOnlyDictionary<TKey, TValue> Instance = System.Collections.Immutable.ImmutableDictionary.Create<TKey, TValue>();
```
slozier, line 146: Could we not do something like for .NET framework?
```C#
internal static ReadOnlyDictionary<TKey, TValue> Instance = new ReadOnlyDictionary<TKey, TValue>(new Dictionary<TKey, TValue>());
```

BCSharp: I have thought about it but `ReadOnlyDictionary` exposes its underlying writable dictionary though a property. So a malicious 3rd party module could navigate to this singleton and inject some keys into a global "empty" instance. Wouldn't it be a security risk?

It could also happen by accident, if the code passes the directory around and at some point, far from the original call place, a bug happens that would cause something fail in some seemingly unrelated place.

I have here locally an implementation of a read-only empty dictionary, it's about 15 lines, we could use that safely.

Now that I think about it: all those singletons should be a `readonly` field or a property with a getter only.

slozier: Well, the `ReadOnlyDictionary<TKey,TValue>.Dictionary` property is `protected` so I'm not particularly worried about accidental modifications (unless there are some technicalities I'm not thinking about).

As for the security aspect I don't know, wouldn't they have to go through reflection? As soon as you throw reflection into the mix what's stopping the malicious user from just replacing the value of `Instance` (unless the runtime enforces the readonly-ness of `readonly`)?

I would be fine with a custom implementation if you think it's worth the effort (although it sounds like the sort of thing I'd want to rip out if I was doing code cleanup :smile:).

Another less efficient and more ugly option would be to make `Instance` a property and return a new dictionary every time on .NET Framework.

Anyway, the whole reason for my suggestion was to try and avoid having `#if NETCOREAPP` in `MakeReadOnlyDictionary` so as long we we can somehow get rid of the conditional I'm happy.

BCSharp: > property is `protected`

OK, it sells it for me. It prevents accidental misuse. I am probably making a fuss about nothing. After all, with the security aspect — that's what `AppDomain` is for.

> I would be fine with a custom implementation if you think it's worth the effort

Not so much an effort question (I have it implemented), but code pollution for a marginal benefit. I would probably want to rip it out the moment I put it in. 😄

slozier: Well, app domains are dead with .NET (Core) so not sure if they help at all.

slozier (approved):

PR #241, opened 2020-10-28
Coerce Extensible<string> to string if used as ParamDictionary key

BCSharp:
This facilitates IronLanguages/ironpython3#999.

It think ideally a given key should be coerced only when `[ParamDictionary]` is applied on a parameter of a generic type with `string` as `TKey`, while maintaining its identity for `object` cases. But this would require much ore extensive changes and so far I haven't see a need for it in IronPython. But the least the DLR can do is to **accept** `Extensible<string>` as the key type.

This coercing could also have been done at the IronPython level, but I believe it is more elegant in the DLR, and, frankly, where it belongs, sine `Extensible<>` is defined here too.

### DIFF
--- a/Src/Microsoft.Dynamic/Actions/DefaultBinder.MethodCalls.cs
+++ b/Src/Microsoft.Dynamic/Actions/DefaultBinder.MethodCalls.cs
@@ -242,12 +242,14 @@ namespace Microsoft.Scripting.Actions {
             Type[] types = testTypes ? new Type[dict.Count] : null;
             int index = 0;
             while (dictEnum.MoveNext()) {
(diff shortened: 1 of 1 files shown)


slozier (approved):
Looks good to me. I assume this will be tested via IronPython (although it would be nice to set up more stand-alone tests at some point).

BCSharp:
Yes, I have tests for this DLR change and other upcoming changes on the IronPython side. I will be submitting them gradually, together with submodule ref updates.

I have looked into adding some tests directly to the DLR test suite, but found no tests testing this part of the system, which I could extend. Maybe not surprising, since parameter handling code in the DLR itself seems unfinished.

PR #244, opened 2020-10-31
Use EmptyArray singleton where useful

BCSharp:
"useful" is here a subjective qualifier.


Src/Microsoft.Scripting/Utils/CollectionExtensions.cs
```diff
@@ -91,8 +91,6 @@ internal static class EmptyReadOnlyCollection<T> {
         internal static readonly ReadOnlyCollection<T> Instance = new ReadOnlyCollection<T>(new T[0]);
     }
 
-    // TODO: Should we use this everywhere for empty arrays?
-    // my thought is, probably more hassle than its worth
     internal static class EmptyArray<T> {
         internal static readonly T[] Instance = new T[0];
```
slozier, line 95: Would there be any value in doing?
```C#
#if NETCOREAPP || NETSTANDARD
        internal static readonly T[] Instance = Array.Empty<T>();
#else
        internal static readonly T[] Instance = new T[0];
#endif
```

BCSharp: Oh, I didn't know that .NET Core has is public. As for the question if it is useful, I am not sure. There is a small benefit if the client code (e.g. IronPython, or an app hosting IronPython) also specializes `Array.Empty` with the same concrete type as the DLR, otherwise there is a small penalty of forcing a specialization of an additional class. So it depends, but either way, I think the difference is negligible.

My primary motivation here was to avoid unnecessarily creating empty arrays on the heap, which then are left around as garbage for the GC to collect. With very few exceptions, I have left the singletons have their own copies (like `EmptyReadOnlyCollection` just above).

slozier: Maybe when we drop support for IronPython 2 we can bump the DLR to .NET Framework 4.6 (or later) and use `Array.Empty<T>()` throughout.

PR #247, opened 2020-11-04
Explicit message for keyword/positional argument conflict

BCSharp:
There are a number of issues with the DLR overload resolver, which I am looking into now, but the code is so wonderfully complex that I prefer to start with something simple, to build confidence. Here is such change: it is not functional (it doesn't change how the resolving happens), it merely makes the error messages more specific in case of duplicate arguments.

Currently, for duplicate arguments, the message is "_MethodName_() got multiple values for keyword argument '_arg_'". To me it suggests that a keyword argument _arg_ is duplicated, but it can also be reported when an argument is provided by position and keyword, which is also a clash.

To be more accurate, positional/keyword clash is not handled well at all, sometimes resulting in strange messages as "_MethodName_() takes at least 2 arguments (6 given)", `IndexOutOfRangeException`, or simply no error but incorrect behaviour. Looking into the code, it seems to me that handling all those cases is not quite finished. But in this PR, I only want to establish what the appropiate error message should be in such cases, and than later build on it when resolving the actual problems.

For reference, Python (CPython) will issue  "_funcname_() got multiple values for keyword argument '_arg_'" **only** when there is a keyword/keyword clash. For positional/keyword clash, CPython uses "_funcname_() got multiple values for argument '_arg_'" (for user defined functions) and "argument for _funcname_() given by name ('_arg_') and position (_n_)" (for built-in functions). Not that DLR has to follow Python, but lacking better ideas I borrowed the builtin message from Python here.

Some notes about the implementation choices: the reason for a specific method binding failure is being carried around in an object of type `CallFailure`. This class has a property `Reason` of enum type `CallFailureReason`. I was tempted to add a new constant to that enum, say, `DuplicateArgument`, but it is in public API, so worth some consideration. On one hand, [.NET Core guidelines](https://docs.microsoft.com/en-us/dotnet/core/compatibility/) allow it, but what would clients do when they read a value that falls outside the expected enum range? Also, [reading comments on the `DuplicateKeyword` contant](https://github.com/IronLanguages/dlr/blob/57dea695e634b221f61b564746ce4aa3c76e66f7/Src/Microsoft.Dynamic/Actions/Calls/CallFailureReason.cs#L24), it seems that the original intent was to cover both cases with the same enum. So I settled on a change that adds an extra property enabling to differentiate the two cases further down the road. Besides, the extra property carries useful information for error reporting. (Although, looking ahead, I see opportunity/need to add more enum values for some new cases). The type of the property could be better `IReadOnlyList<int>`, but that is also true for the other like-like properties, so I stayed with `List<int>` for consistency. 

Arguably, the error message could have simply been "_MethodName_() got multiple values for argument '_arg_'", but since the information abouth the positional argument is there, why not show it to the user. It will either help the user to diagnose their problem (if the reported position correct), or help us to spot bugs in the DLR (if incorrect).

This brings me to the last point: about the meaning of the _position_. As I see it, it can be interpreted as either a position of the offending argument that is given without a name, or a position of the parameter, which is being claimed by a positional argument and a keyword argument. Typically they are the same, but not always. Personally, as a user, I would expect the position of the argument being reported, but the DLR reports the position of the parameter. Also, interestingly, CPython reports the position of the parameter, so I assume this is fine. Here is an example:

~~~
Python 3.4.4 (v3.4.4:737efcadf5a6, Dec 20 2015, 20:20:57) [MSC v.1600 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import datetime
>>> datetime.date(year=2020, *(11, 2)) 
Traceback (most recent call last):   
  File "<stdin>", line 1, in <module>
TypeError: Argument given by name ('year') and position (1)
~~~

### DIFF
--- a/Src/Microsoft.Dynamic/Actions/Calls/ActualArguments.cs
+++ b/Src/Microsoft.Dynamic/Actions/Calls/ActualArguments.cs
@@ -92,6 +92,7 @@ namespace Microsoft.Scripting.Actions.Calls {
 
             List<string> unboundNames = null;
             List<string> duppedNames = null;
+            List<int> duppedPositionals = null;
 
             int positionalArgCount = Arguments.Count;
 
@@ -100,20 +101,23 @@ namespace Microsoft.Scripting.Actions.Calls {
                 if (paramIndex >= 0) {
                     int nameIndex = paramIndex - positionalArgCount;
 
-                    // argument maps to already bound parameter:
-                    if (paramIndex < positionalArgCount || boundParameters[nameIndex]) {
-                        if (duppedNames == null) {
-                            duppedNames = new List<string>();
-                        }
+                    if (paramIndex < positionalArgCount) {
+                        // argument maps to already bound positional parameter
+                        duppedNames ??= new List<string>();
+                        duppedPositionals ??= new List<int>(duppedNames.Count);
                         duppedNames.Add(ArgNames[i]);
+                        duppedPositionals.Add(method.PositionOfParameter(ArgNames[i]));
+                    } else if (boundParameters[nameIndex]) {
+                        // argument maps to already bound named parameter
+                        duppedNames ??= new List<string>();
+                        duppedNames.Add(ArgNames[i]);
+                        duppedPositionals?.Add(0);
                     } else {
                         permutation[i] = nameIndex;
                         boundParameters[nameIndex] = true;
                     }
                 } else {
-                    if (unboundNames == null) {
-                        unboundNames = new List<string>();
-                    }
+                    unboundNames ??= new List<string>();
                     unboundNames.Add(ArgNames[i]);
                 }
             }
@@ -121,12 +125,14 @@ namespace Microsoft.Scripting.Actions.Calls {
             binding = new ArgumentBinding(positionalArgCount, permutation);
 
             if (unboundNames != null) {
-                failure = new CallFailure(method, unboundNames.ToArray(), true);
+                failure = new CallFailure(method, unboundNames.ToArray(), unassignable: true);
                 return false;
             }
 
             if (duppedNames != null) {
-                failure = new CallFailure(method, duppedNames.ToArray(), false);
+                failure = duppedPositionals != null ?
+                    new CallFailure(method, duppedNames.ToArray(), duppedPositionals.ToArray()) :
+                    new CallFailure(method, duppedNames.ToArray(), unassignable: false);
                 return false;
             }
 
--- a/Src/Microsoft.Dynamic/Actions/Calls/ActualArguments.cs
+++ b/Src/Microsoft.Dynamic/Actions/Calls/ActualArguments.cs
@@ -4,7 +4,9 @@
 
 using System.Collections;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Dynamic;
+using System.Linq;
 
 using Microsoft.Scripting.Utils;
 
@@ -104,7 +106,7 @@ namespace Microsoft.Scripting.Actions.Calls {
                     if (paramIndex < positionalArgCount) {
                         // argument maps to already bound positional parameter
                         duppedNames ??= new List<string>();
-                        duppedPositionals ??= new List<int>(duppedNames.Count);
+                        duppedPositionals ??= Enumerable.Repeat(0, duppedNames.Count).ToList();
                         duppedNames.Add(ArgNames[i]);
                         duppedPositionals.Add(method.PositionOfParameter(ArgNames[i]));
                     } else if (boundParameters[nameIndex]) {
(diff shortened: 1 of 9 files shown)


Src/Microsoft.Dynamic/Actions/Calls/ActualArguments.cs
```diff
@@ -100,33 +101,38 @@ public sealed class ActualArguments {
...
+                        // argument maps to already bound positional parameter
+                        duppedNames ??= new List<string>();
+                        duppedPositionals ??= new List<int>(duppedNames.Count);
                         duppedNames.Add(ArgNames[i]);
+                        duppedPositionals.Add(method.PositionOfParameter(ArgNames[i]));
+                    } else if (boundParameters[nameIndex]) {
+                        // argument maps to already bound named parameter
+                        duppedNames ??= new List<string>();
+                        duppedNames.Add(ArgNames[i]);
+                        duppedPositionals?.Add(0);
```
slozier, line 114: Is it possible for the two lists to go out of sync? For example if this branch is taken first and then the other branch? If not then wouldn't the following hold true?
```C#
Debug.Assert(duppedPositionals is null || duppedNames.Length == duppedPositionals.Length);
```

Src/Microsoft.Dynamic/Actions/Calls/MethodCandidate.cs
```diff
@@ -95,6 +95,22 @@ public sealed class MethodCandidate {
             return -1;
         }
 
+        internal int PositionOfParameter(string name) {
+            for (int i = 0, p = 1; i < _parameters.Count; i++) {
+                if (_parameters[i].IsHidden) {
```
slozier, line 100: Maybe:
```C#
var parameter = _parameters[i];
```

Src/Microsoft.Dynamic/Runtime/BinderOps.cs
```diff
@@ -152,6 +152,10 @@ public static class BinderOps {
             return new ArgumentTypeException($"{name}() got multiple values for keyword argument '{argumentName}'");
         }
 
+        public static ArgumentTypeException TypeErrorForDuplicateArgument(string name, int position, string argumentName) {
+            return new ArgumentTypeException($"Argument for {name}() given by name ('{argumentName}') and position ({position})");
```
slozier, line 156: If we're picking the message to match CPython, then in newer versions it looks like they use lower-case "a" in argument. I don't particularly care either way, just something to consider.

Src/Microsoft.Dynamic/Actions/Calls/CallFailure.cs
```diff
@@ -17,27 +20,41 @@ namespace Microsoft.Scripting.Actions.Calls {
...
         }
 
         internal CallFailure(MethodCandidate candidate, string[] keywordArgs, bool unassignable) {
+            Candidate = candidate;
             Reason = unassignable ? CallFailureReason.UnassignableKeyword : CallFailureReason.DuplicateKeyword;
+            _keywordArgs = keywordArgs;
+            _positionalArgs = EmptyArray<int>.Instance;
+        }
+
+        internal CallFailure(MethodCandidate candidate, string[] keywordArgs, int[] positionalArgs) {
```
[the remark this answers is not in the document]
BCSharp: I like your suggestion. Luckily it is internal, I wouldn't want to change a public constructor.

Src/Microsoft.Dynamic/Runtime/BinderOps.cs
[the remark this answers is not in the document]
BCSharp: I was aware of that. I didn't elaborate that in my initial description because it was already long and convoluted...

I chose the capital "A" because it is the DLR and all (most?) messages it produces are properly sentence-capitalized. This is also the convention throughout .NET. I don't particularly care either, as long as the messages are clear, unambiguous, fairly precise, helpful, and accurate. Also, the same error condition can be perceived and reported from various angles. For instance, what if there is a positional, and two keyword arguments clashing? I think it doesn't matter as long as the message is correct from some point.

CPython is also not consistent in its choices of reporting perspectives. As a teaser, the example I gave in my PR description looks differently in 3.9:

~~~
Python 3.9.0 (tags/v3.9.0:9cf6752, Oct  5 2020, 15:34:40) [MSC v.1927 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import datetime
>>> datetime.date(year=2020, *(11, 2))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: function missing required argument 'day' (pos 3)
~~~

Src/Microsoft.Dynamic/Actions/Calls/ActualArguments.cs
BCSharp: My mistake. I forgot that the number given to the List constructor is the capacity, not size. The assertion you wrote was supposed to hold. Although strictly speaking `duppedPositionals` are allowed to be shorter than `duppedNames`, this is not what the code here intends to do.

Thanks for checking!

slozier (approved):
Thanks!

PR #248, opened 2020-11-07
Use `nameof` for static names

BCSharp:
This assists static code analyzers like Visual Studio, and ultimately efforts to understand/reverse-engineer the code.

The update is extensive but fairly straightforward, and since there is no functional change I hope i can qualify to be included before release. There will be an accompanying update on the IronPython 3 side because of a change to a code generation script.


slozier (approved):
Looks good to me. Any reason why you used the full namespace for the `Marshal` nameofs?

BCSharp:
In some classes the type `Marshal` clashes with the same-named method overridden in those classes.

Issue #249, opened 2020-11-13
A case against positional or by keyword binding of ParamArray/ParamDictionary parameters


BCSharp:
There is a certain inconsistency in handling parameters marked with `ParamArrayAttribute` or `ParamDictionaryAttribute`. It is reflected in both comments in the DLR code, as well as code itself. Since this is a fundamental concept in the overload resolution, I want to discuss it explicitly and agree on a principle, before changing the DLR behaviour in one way or another. It would be very useful to know which behaviour is desirable and which should be considered a bug. Once this principle is agreed on and implemented, it may be difficult to reverse on it in the future, since many parts of the overload resolution code depend on it.

Let's look at the current situation first. For instance, `ParameterWrapper.cs` [reads](https://github.com/IronLanguages/dlr/blob/af4d5595d593596b64cf2c0337de2fd8ae1bf47b/Src/Microsoft.Dynamic/Actions/Calls/ParameterWrapper.cs#L60):

    // params arrays & dictionaries don't allow assignment by keyword

On the other hand, a comment on `ParamDictionaryAttribute` [reads](https://github.com/IronLanguages/dlr/blob/af4d5595d593596b64cf2c0337de2fd8ae1bf47b/Src/Microsoft.Scripting/Runtime/ParamDictionaryAttribute.cs#L24):

    /// For languages which don't have language level support the user will be required to
    /// create and populate the dictionary by hand.
    /// 
    /// This attribute is the dictionary equivalent of the System.ParamArrayAttribute.

This suggests that it should be possible to pass such param dictionary and/or param array somehow. Since passing by keyword should be forbidden, this leaves that passing by position should be supported. But this leads to logical inconsistencies, which I describe a little bit later below.

On the actual code side, some parts of the code assume that passing param dictionary/array by keyword is allowed, and some assume that it is forbidden. This leads to erroneous behaviour, ranging from unexpected results, through confusing error messages, to exceptions thrown. Example (from IronPython):

```Python
import clr
clr.AddReference("rowantest.methodargs")
from Merlin.Testing.Call import VariousParameters
obj = VariousParameters()
obj.M560(1, 2, 3, z=4, b=5, c=6)
```
Method `M560` has signature `public void M560(int x, [ParamDictionary] IDictionary<string, int> y, params int[] z)` and the call should succeed. Instead, and `IndexOutOfRangeException` is thrown:
```
Traceback (most recent call last):   
  File "<stdin>", line 1, in <module>
[...]

PR #250, opened 2020-11-17
Prevent ParamArray and ParamDictionary parameters from binding by keyword

BCSharp:
This fixes

```Python
import clr
clr.AddReference("rowantest.methodargs")
from Merlin.Testing.Call import VariousParameters
obj = VariousParameters()
obj.M560(1, 2, 3, z=4, b=5, c=6)
```

reported in #249.

Disabling by-keyword assignment to param arrays and dictionaries is controlled though a virtual method on `OverloadResolver`, but I stopped short from making it protected since the DLR currently cannot handle the "enabled" case so I did not want the languages to start requesting it. It can go in three ways from here:

1. The DLR implements `PositonalOnlyAttribute`/`NonKeywordAttribute` for the benefit of all languages. Probably the best way to support it is in `ReflectionUtils` and the virtual extension point maybe can be abolished (i.e. preventing binding by keyword to param arrays/dictionaries will be hard-coded).
2. `PositonalOnlyAttribute`/`NonKeywordAttribute` will be implemented at the language level (i.e. in IronPython), and the virtual extension point will become protected and overridden in `PythonOverloadResolver`, (or any other language-specific resolver derived from `DefaultOverloadResolver`, with the provision that the language can only tighten the restriction rather than relax it).
3. The DLR will somehow be upgraded/cleaned/up/extended to be able to handle param arrays/dictionaries being bound by keyword, by some to be determined consistent set of rules. Then the virtual extension point can become protected without any restrictions.

### DIFF
--- a/Src/Microsoft.Dynamic/Actions/Calls/DefaultOverloadResolver.cs
+++ b/Src/Microsoft.Dynamic/Actions/Calls/DefaultOverloadResolver.cs
@@ -139,9 +139,10 @@ namespace Microsoft.Scripting.Actions {
             }
         }
 
-        // TODO: review the signature, add protected
-        internal override bool AllowByKeywordArgument(OverloadInfo method, ParameterInfo parameter) {
-            return !parameter.ProhibitsByKeyword();
+        protected internal override bool AllowByKeywordArgument(OverloadInfo method, ParameterInfo parameter) {
+            // params arrays & dictionaries don't allow assignment by keyword
+            return base.AllowByKeywordArgument(method, parameter)
+                && !parameter.IsParamArray() && !parameter.IsParamDictionary();
         }
 
         protected override ActualArguments CreateActualArguments(IList<DynamicMetaObject> namedArgs, IList<string> argNames, int preSplatLimit, int postSplatLimit) {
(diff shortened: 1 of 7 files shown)


BCSharp:
After sleeping on it, I now think that option 2 is the best way forward. Implementing new attributes in IronPython 3 will allow to get the feeling how they work and possibly tweak things before releasing them to the world. 

@slozier if this is OK with you, I will amend the PR accordingly.

slozier:
> if this is OK with you, I will amend the PR accordingly.

Sounds good to me!

slozier (approved):
Looks good to me. Causes some failures in ipy3 but I believe they are expected.

Are you going to have to remove internal from some of the virtual methods to get this working from ipy3?

BCSharp:
> Causes some failures in ipy3 but I believe they are expected.

I have updates to ipy3 tests locally ready to go once an official DLR commit lands on master.

> Are you going to have to remove internal from some of the virtual methods to get PositonalOnlyAttribute/NonKeywordAttribute working from ipy3?

No, all virtual methods I foresee needing to be overridden in `PythonOverloadResolver` are already `protected` or `protected internal`. I haven't looked into KeywordOnlyAttribute/NonPositionalAttribute yet so perhaps some DLR update will be needed for that, but I am parking this story for now to look into nonlocal in ipy3.

PR #254, opened 2021-09-05
Support for implementation specific options using -X

slozier:
Related to https://github.com/IronLanguages/ironpython3/issues/1194


slozier:
Probably worth looking the other option related issues at the same time https://github.com/IronLanguages/ironpython3/issues/1075 and https://github.com/IronLanguages/ironpython3/issues/1074

PR #256, opened 2022-02-22
BlazorWasm support

TwoUnderscorez:
When trying to use IronPython3 from a Blazor web-assembly application, `Microsoft.Scripting.Runtime.SharedIO` throws an exception when trying to setup `stdin`, which is understandable, because there is none.

All I did is catch the exception and set the stream to `Stream.Null`.

### DIFF
--- a/Src/Microsoft.Scripting/Runtime/SharedIO.cs
+++ b/Src/Microsoft.Scripting/Runtime/SharedIO.cs
@@ -105,7 +105,7 @@ namespace Microsoft.Scripting.Runtime {
                         _inputStream = Stream.Null;
                         _inputReader = TextReader.Null;
 #endif
-                        } catch (TypeInitializationException) {
+                        } catch (PlatformNotSupportedException) {
                             // When on BlazorWebassembly, this exception is thrown.
                             _inputEncoding = Encoding.UTF8;
                             _inputStream = Stream.Null;
--- a/Src/Microsoft.Scripting/Runtime/SharedIO.cs
+++ b/Src/Microsoft.Scripting/Runtime/SharedIO.cs
@@ -91,10 +91,11 @@ namespace Microsoft.Scripting.Runtime {
             if (_inputStream == null) {
                 lock (_mutex) {
                     if (_inputStream == null) {
+                        try {
 #if FEATURE_FULL_CONSOLE
-                        _inputStream = ConsoleInputStream.Instance;
-                        _inputEncoding = Console.InputEncoding;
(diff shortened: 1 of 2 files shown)


slozier (changes requested):
Thanks for the PR. My main concern with this approach is that if someone changes the order of the assignments (I think) it'll throw a PNSE instead of a `TypeInitializationException`. Probably better to fix `ConsoleInputStream` to not throw on initialization and then catch the PNSE.

I guess my other problem is I don't have a Blazor test rig to validate this. Any pointers on setting one up?

Related issue: https://github.com/IronLanguages/ironpython2/issues/769
Workaround: https://github.com/IronLanguages/ironpython2/issues/769#issuecomment-767590643

TwoUnderscorez:
@slozier Is that better?

slozier (approved):
Thanks, that works.

Going from a field to a property is a breaking change but I assume no one outside the assembly is using this... Also I think we lose thread safety on the singleton creation. Will need to look into it a bit more, but I guess I can fix it after the merge if required.

TwoUnderscorez:
@slozier Sorry, I don't quite understand how this is a breaking change. If you would point me in the right direction I might be able to make it a non-breaking change.

slozier:
@TwoUnderscorez I meant having `ConsoleInputStream.Instance` go from a field to a property is technically a binary breaking change (for example if assembly A uses a field from assembly B and you replace B with B' which contains a property instead, then A is broken). I don't think I'm really worried about breaking anyone with this particular scenario since the only consumer is in the same assembly - but it is a public type so it's not impossible...

An alternative approach would be to leave `ConsoleInputStream.Instance` as a field and change the constructor to not throw:
```C#
private ConsoleInputStream() {
    try {
        _input = Console.OpenStandardInput();
    } catch (PlatformNotSupportedException) {
        _input = Stream.Null;
    }
}
```

TwoUnderscorez:
@slozier I could do that, but I would have to leave the catch block in `ShareIO.InitializeInput` because 
```csharp
_inputEncoding = Console.InputEncoding;
_inputReader = Console.In;
``` 
also throw `PlatformNotSupportedException` (maybe only one of them), which I personally think is kind of ugly. Would you like me to do what you've suggested or merge as is currently?

slozier:
@TwoUnderscorez I think the PNSE try/catch you have in `InitializeInput` is fine, but I'd change the `ConsoleInputStream` to what I suggested instead.

slozier:
Thanks for the PR!

PR #268, opened 2023-07-04
Skip double FindEntry in ScopedStorage.TryGetValue

arontsang:
Fast path when item is found in ScopedStorage, skips redundant FindEntry call


slozier (approved):
Thanks, looks good to me.

arontsang:
Thanks, that was super quick.

PR #270, opened 2023-07-25
Resurrect TextStream

BCSharp:
`TextStream` was removed as part of Silverlight cleanup (#25) but is still required to build the DLR if `FEATURE_FULL_CONSOLE` is disabled but `FEATURE_BASIC_CONSOLE` enabled.

Besides, it is useful to run IronPython 3 in LinqPad.

### DIFF
--- a/Src/Microsoft.Scripting/Utils/TextStream.cs
+++ b/Src/Microsoft.Scripting/Utils/TextStream.cs
@@ -1,0 +1,140 @@
+/* ****************************************************************************
+ *
+ * Copyright (c) Microsoft Corporation. 
+ *
+ * This source code is subject to terms and conditions of the Apache License, Version 2.0. A 
+ * copy of the license can be found in the License.html file at the root of this distribution. If 
+ * you cannot locate the  Apache License, Version 2.0, please send an email to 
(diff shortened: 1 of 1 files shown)


slozier (approved):
Fine by me. Surprising number of people seem to have that stream/writer issue. Wonder if there's anything we could do on our end to help...

BCSharp:
I suppose "help" here could mean three things:
1. Do something that the issue becomes transparent (does not appear) at least in some if not all cases.
2. Make sure that people are aware of the issue and know how to handle it.
3. Make is easier/simpler to handle the issue.

Ad 1: I have no idea.
Ad 2: Maybe moving the section about this issue to the top of `upgrading-from-ipy2`. It's a detailed document so maybe not every reader reaches to the end.
Ad 3: Bringing `TextStream` back goes in this direction. I am still thinking about other possibilities.

slozier:
I don't know if I would make it transparent even if I could figure out how to do it, but if we could offer some sort of opt-in mechanism or a one-liner it would probably be helpful. Anyway, as you said, bringing `TextStream` back is a step in that direction...

PR #271, opened 2023-07-31
Fix option ColorfulConsole

BCSharp:
Option `-X ColorfulConsole` was not working, for several reasons:

1. The fag mask used to set colors was incorrect, resulting all colors be coerced to either white or black.
2. The fallback color was white, which for white background always resulted in the text being invisible (macOS default console has white background).
3. The background detection was working only on Windows, on Unix, the background color is unknown. So to make this option usable on Unix systems, I extended it with an optional parameter of value _dark_ or _light_. To prevent proliferation of constructor parameters and the number of constructors of `BasicConsole` and `SuperConsole`, they both got a constructor that accepts `ConsoleOptions` and picks up the option it can understand.

### DIFF
--- a/Src/Microsoft.Dynamic/Hosting/Shell/OptionsParser.cs
+++ b/Src/Microsoft.Dynamic/Hosting/Shell/OptionsParser.cs
@@ -180,7 +180,6 @@ namespace Microsoft.Scripting.Hosting.Shell {
                 case "-X:PrivateBinding":
                 case "-X:PassExceptions":
                 // TODO: #if !IRONPYTHON_WINDOW
-                case "-X:ColorfulConsole":
                 case "-X:TabCompletion":
                 case "-X:AutoIndent":
                 // #endif
@@ -197,6 +196,7 @@ namespace Microsoft.Scripting.Hosting.Shell {
 #if DEBUG
                 case "-X:PerfStats":
 #endif
+                    // Option without argument
                     HandleImplementationSpecificOption(arg.Substring(3), null);
                     break;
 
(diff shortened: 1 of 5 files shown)


Src/Microsoft.Dynamic/Hosting/Shell/OptionsParser.cs
```diff
@@ -327,7 +333,8 @@ public class OptionsParser<TConsoleOptions> : OptionsParser
                 { "-X:PrivateBinding",           "Enable binding to private members" },
                 { "-X:ShowClrExceptions",        "Display CLS Exception information" },
                 { "-X:TabCompletion",            "Enable TabCompletion mode" },
-                { "-X:ColorfulConsole",          "Enable ColorfulConsole" },
+                { "-X:ColorfulConsole[=(dark|light)]",
```
slozier, line 336: Does using `=` instead of a space work in both 2 and 3? I know I changed 3 to more closely resemble CPython but I forget the details (and am unable to test right now). Maybe not a big deal since I don't plan on another release of 2...

slozier (approved):

Src/Microsoft.Dynamic/Hosting/Shell/OptionsParser.cs
BCSharp: It cannot be a space iso `=` here because the argument is optional. With a space it would be mistaken for a file argument.

This will not work with 2 (which BTW is not buildable right now because the DLR is on net452). Also, `-X:ColorfulConsole=light` (with `:` iso space) does not work with 3. This by itself is not a problem (since there is no backward compatibility need here nor is this form shown in the help printout), but it kind of makes the DLR not self-consistent and unable to process the optional argument at all. It happens to work with 3 because of all the extra tweaks that it does to options before handing them over to the base class in the DLR for processing.

I can add support for this form to the DLR, just for options with an optional argument (of which this is the first one). An alternative could be to modify the DLR that any mandatory option argument can be given separated by a space ***or*** `=`. The DLR has between 1 and 3 of those (depending on a feature set). What do you think?

slozier: I'm having a hard time making up my mind because I don't know who (if anyone) outside of us is using the cli parsing. If we allow both spaces and equal on mandatory arguments I think we'd be inconsistent with Python? Probably easiest to add support for this form. Could even just be hard coded in the switch statement (e.g. `-X:ColorfulConsole=true`)...

slozier (approved):
Thanks!

PR #272, opened 2023-08-13
Make BASIC and FULL console support runtime-selectable if both available

BCSharp:
The level of console support is selectable at build time through compile symbols `FEATURE_FULL_CONSOLE` and `FEATURE_BASIC_CONSOLE`.

1. When `FEATURE_FULL_CONSOLE` is defined, standard IO is performed using .NET console streams. This option is preferable for a standalone `ipy` program.
2. When `FEATURE_BASIC_CONSOLE` is defined, standard IO is performed using console reader/writer. This option may be preferable in some embedded cases when the hosting source code is not available and performs its own redirection (like _LinqPad_).
3. When neither is defined, standard IO is silently discarded.
4. When both are defined at the same time, the current code defaults to using .NET console streams.

The last case is interesting because this is the build feature set that is available on NuGet. The change proposed here makes the selection between using console streams and console reader/writer available at runtime, in case both relevant compile symbols were defined during the build. In this way, IronPython from NuGet could be configured for the embedded scenario in a simple way, like with a runtime option.

### DIFF
--- a/Src/Microsoft.Scripting/Runtime/SharedIO.cs
+++ b/Src/Microsoft.Scripting/Runtime/SharedIO.cs
@@ -84,6 +84,10 @@ namespace Microsoft.Scripting.Runtime {
         public Encoding OutputEncoding { get { InitializeOutput(); return _outputWriter.Encoding; } }
         public Encoding ErrorEncoding { get { InitializeErrorOutput(); return _errorWriter.Encoding; } }
 
+#if FEATURE_FULL_CONSOLE && FEATURE_BASIC_CONSOLE
+        public bool PreferBasicIO { get; set; }
+#endif
+
         internal SharedIO() {
         }
 
@@ -93,14 +97,22 @@ namespace Microsoft.Scripting.Runtime {
                     if (_inputStream == null) {
                         try {
 #if FEATURE_FULL_CONSOLE
-                            _inputStream = ConsoleInputStream.Instance;
-                            _inputEncoding = Console.InputEncoding;
-                            _inputReader = Console.In;
-#elif FEATURE_BASIC_CONSOLE
+    #if FEATURE_BASIC_CONSOLE
+                            if (!PreferBasicIO) {
+    #endif
+                                _inputEncoding = Console.InputEncoding;
+                                _inputStream = ConsoleInputStream.Instance;
+                                _inputReader = Console.In;
+    #if FEATURE_BASIC_CONSOLE
+                                return;
+                            }
+    #endif
+#endif
+#if FEATURE_BASIC_CONSOLE
                             _inputEncoding = Encoding.Default;
                             _inputStream = new TextStream(Console.In, _inputEncoding);
                             _inputReader = Console.In;
-#else
+#elif !FEATURE_FULL_CONSOLE
                             _inputEncoding = Encoding.UTF8;
                             _inputStream = Stream.Null;
                             _inputReader = TextReader.Null;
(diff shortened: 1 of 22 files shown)


slozier:
What about getting rid of the feature flags? As far as I can tell, their whole reason for being was to support silverlight/win8 builds which had limited API surfaces. With the .NET standard efforts I think the APIs are pretty much guaranteed to exist on anything .NET (although they could PNSE).

Also, if we're going to be exposing this feature I wonder if we should be using a different encoding (e.g. not `Encoding.Default`).

BCSharp:
> What about getting rid of the feature flags? 

If you are OK with it, removing these flags would be even better. In such case I would make it runtime-selectable between the three versions currently supported by the compile switches. In this way any project that currently depends on them could easily migrate.

> Also, if we're going to be exposing this feature I wonder if we should be using a different encoding (e.g. not `Encoding.Default`).

I'll have to look into it (run some tests etc.)

Src/Microsoft.Scripting/Runtime/SharedIO.cs
```diff
@@ -132,24 +127,21 @@ private sealed class StreamProxy : Stream {
...
+                                _outputWriter = Console.Out;
+                                break;
+                            case SupportLevel.Basic:
+                                _outputStream = new TextStream(Console.Out);
+                                _outputWriter = Console.Out;
+                                break;
+                            case SupportLevel.None:
+                            default:
+                                _outputStream = Stream.Null;
+                                _outputWriter = new StreamWriter(_outputStream, Encoding.UTF8, bufferSize: 16);
```
slozier, line 142: The `StreamWriter` instead of `TextWriter.Null` is to have an encoding check?

slozier (approved):
Looks good to me. Definitely cleaner without the flags. Just have the one question but feel free to merge otherwise.

Src/Microsoft.Scripting/Runtime/SharedIO.cs
BCSharp: `TextWriter.Null` on .NET Framework uses a codepage encoding (probably `Encoding.Default`), so printing from ipy throws exceptions when the text to print contains characters that cannot be encoded in that codepage.

.NET, on the other hand, uses `Encoding.Unicode`, which can encode anything. I will modify the code so that it uses `StreamWriter` only when necessary.

slozier: Hmm, interesting. Was assuming `TextWriter.Null` would just send stuff to the ether. Reference implementation (see below) itself (for .NET Framework) doesn't seem to make any use encoding so maybe we're the ones using it...
```cs
        [Serializable]
        private sealed class NullTextWriter : TextWriter
        {
            internal NullTextWriter(): base(CultureInfo.InvariantCulture) {
            }
        
            public override Encoding Encoding {
                get { return Encoding.Default;  }
            }
 
            [SuppressMessage("Microsoft.Contracts", "CC1055")]  // Skip extra error checking to avoid *potential* AppCompat problems.
            public override void Write(char[] buffer, int index, int count) {
            }
            
            public override void Write(String value) {
            }
 
            // Not strictly necessary, but for perf reasons
            public override void WriteLine() {
            }
 
            // Not strictly necessary, but for perf reasons
            public override void WriteLine(String value) {
            }
            
            public override void WriteLine(Object value) {
            }
        }
```

BCSharp: Yes, `TextWriter.Null` swallows anything without problem. It is IronPython that complaints. I haven't investigated where specifically it happens, just observed the exception.

slozier: > Is it possible that IronPython uses the encoding of the output writer to encode the text before sending it to the output stream directly?

Ah right, the `OutputEncoding` returns `_outputWriter.Encoding` and that's used to wrap the console stream in ipy.

BCSharp: > Edit: kind of bad if you mutate your output after that's created.

Do you mean change `ConsoleSupportLevel` and call `RedirectToConsole()` after stdio files are created? In ipy, we can make sure it does not happen, and `SharedIO` access is [reserved to hosting providers like ipy](https://github.com/IronLanguages/dlr/blob/1767f9c5e6f1d168ebc9b91aa657e5af99e2588a/Src/Microsoft.Scripting/Hosting/Providers/HostingHelpers.cs#L18-L22).

slozier: I guess I was thinking about the embedded scenario where people are having issues because they change the input/output  after their engine is created (whether done via `RedirectToConsole` or otherwise) and stdio encoding is locked down in the `TextIOWrapper`.

Anyway, not an issue with this - just something that popped to mind when I looked at the usage. Might be related to the recent email on the user group where someone called `_Engine.Runtime.IO.SetOutput(MyStream, Encoding.UTF8)` and print started outputting chinese characters.

BCSharp:
> feel free to merge

I don't see how I could do it. Do I have write access to this repo?

Src/Microsoft.Scripting/Runtime/SharedIO.cs
BCSharp: OK, thanks, it makes sense.

slozier:
Ah hmm, thought you had access! Should work now.

PR #274, opened 2023-08-29
Fix tests on latest preview SDK

slozier:


slozier:
Seems like it's not actually running net452 and netcoreapp2.1...

BCSharp:
`net462` not packaged anymore?

slozier:
> `net462` not packaged anymore?

Retargeting the test projects to `net462` started generating `net462` assemblies which did not exist before, so excluding them is just maintains the status quo. Although I'm experimenting with bumping everything to `net462` and dropping `netcoreapp3.1`. This means we would also drop testing on .NET Core 2.1 and use 3.1 to exercise the `netstandard2.0` assemblies...

Regarding the macOS failures it will supposedly be fixed with the 7.0.401 SDK so we just have to wait a while for it to make its way over to the agents...

PR #279, opened 2023-12-10
Avoid ToLoverInvariant in string comparisons (CA1862)

BCSharp:
Resolves some warnings-turned-errors.


slozier (commented):
Seems fine to me although I'm thinking it would be better if these were `OrdinalIgnoreCase`?

BCSharp:
I suppose you're right, `OrdinalIgnoreCase` should be faster.

PR #281, opened 2024-11-27
Fix for missing trace event on first expression

slozier:
This should resolve the issue from https://github.com/IronLanguages/ironpython3/pull/1810 where the first trace event is not getting generated.


BCSharp:
I suppose the build/packaging fixes will go through a separate PR?

slozier:
> I suppose the build/packaging fixes will go through a separate PR?

I ended up doing a push to `main` (yes I know it's bad) for the revert instead of going through the github revert mechanism. The build/packaging fixes should be on the `main` branch.

PR #329, opened 2026-04-14
Reactivate HostingTest

BCSharp:
This is an old test. It wasn't passing, it wasn't even compiling. The copyright message in the assembly attribute read 2008. That is 18 years ago. My goal here was to get it compilable and passable again, with limited changes.

The main change is the migration to the SDK format for the project file and upgrade NUnit to v.3. The project was only targeting .NET Framework and I kept it so, mainly due to remoting tests. I think that many tests not related to remoting could be run on .NET, but it requires some more rework.

Perhaps the project was active before the main repo split because it assumes IronPython two levels up; or around this time. I don't mind IronPython depending on the DLR submodule, but I don't think it is a good idea for the DLR do depend on the IronPython souce code. Maybe it can be modified to depend on IronPython from NuGet; it is not supposed to test IronPython, but the DLR.

The project also assumed IronRuby to be around. I modified it to skip IronRuby-dependent tests if IronRuby is not present, which, in my case, is always so.

There is one test that looked suspicious to me and not passing, so I disabled (ignored) it. There are more originally ignored tests, indicating some sort of bugs. I left them as they were.

There is more work to be done on this project, like removal of warnings, new array initialization syntax, etc.

The test is not part of the standard test suite for the DLR, nor is it run by CI. It might be enabled by CI of IronPython, but then it is guarding the wrong place. Maybe if it is made depending on a NuGet package, it can be enabled on the DLR CI. For now, run it locally (Windows only) within an IronPython repo by:

```
dotnet test .\src\dlr\tests\HostingTest
```


slozier (approved):
Thanks! DLR could definitely use more tests.

If we wanted to run this during CI, I guess pulling ipy from nuget would be fine...

PR #334, opened 2026-04-16
Simplify usage of ContainsKey

BCSharp:
Core source code only scanned.


src/core/Microsoft.Dynamic/Actions/NamespaceTracker.cs
```diff
@@ -109,12 +109,11 @@ internal void AddTypeName(string typeName, Assembly assem) {
             typeNames.AddTypeName(typeName);
 
             string normalizedTypeName = ReflectionUtils.GetNormalizedTypeName(typeName);
-            if (_dict.ContainsKey(normalizedTypeName)) {
+            if (_dict.TryGetValue(normalizedTypeName, out MemberTracker existingValue)) {
```
slozier, line 112: I thought there was an analyzer for this? Maybe the pattern is too complex for it...

slozier (approved):

src/core/Microsoft.Dynamic/Actions/NamespaceTracker.cs
BCSharp: Yes, there is: CA1854, set as _suggestion_, the same as in IronPython. I think you are right, the pattern was too complex to be picked up, even as _error_. Nevertheless, shall we set it as _warning_?

slozier: Warning works for me.

slozier (approved):

PR #335, opened 2026-04-16
Correct indentation of Python snippets

BCSharp:
Syntactic sugar only, to make IDE editors properly guess indentation used. Ruby snippets are also messed up but they seem to be a copy of Python? In any case, they (or not all or them) are Ruby, so I am not touching it. Ruby tests are not performed.

### DIFF
--- a/tests/HostingTest/PreDefinedCodeSnippets.cs
+++ b/tests/HostingTest/PreDefinedCodeSnippets.cs
@@ -6,33 +6,20 @@ using System;
 
 namespace HostingTest {
 
-    public class CodeSnippet {
-        public string ID;
-        public string Code;
-        public string Description;
-
-        public CodeSnippet(string id, string description, string code) {
-            ID = id ; Code = code;
-            Description = description;
-        }
+    public class CodeSnippet(string id, string description, string code)
+    {
+        public string ID = id;
+        public string Code = code;
+        public string Description = description;
     }
-    internal class CodeSnippetCollection {
-        internal CodeSnippet[] AllSnippets = null;
-        internal CodeSnippet GetCodeSnippetByID(string id) {
-            foreach (CodeSnippet cs in AllSnippets) {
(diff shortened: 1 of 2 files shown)


tests/HostingTest/PythonCodeSnippets.cs
```diff
@@ -4,235 +4,253 @@
...
+                    CodeType.IsEvenFunction, "A function that returns true or false depending on if a number is even or not",
+                    "def iseven(n): return 1 != n % 2"),
+
+                new CodeSnippet(
+                    CodeType.IsOddFunction, "A function that returns true or false depending on if a number is odd or not",
+                    "def isodd(n): return 1 == n % 2;"),
+
+                new CodeSnippet(
+                    CodeType.MethodWithDocumentationAttached, "A very simple method with docs attached",
+                    """"
```
slozier, line 227: Four quotes?! That's a new one for me. :smile:

tests/HostingTest/PreDefinedCodeSnippets.cs
```diff
@@ -6,66 +6,41 @@
...
-        public string ID;
-        public string Code;
-        public string Description;
-
-        public CodeSnippet(string id, string description, string code) {
-            ID = id ; Code = code;
-            Description = description;
-        }
+    public class CodeSnippet(string id, string description, string code)
+    {
```
slozier: nit: { on new line not consistent with project code style

slozier (approved):
Surprised this all works with indents like that. As long as it works I guess.

tests/HostingTest/PythonCodeSnippets.cs
BCSharp: Any number that works for you 😄 (at least three). In this case, three were already within the string.

tests/HostingTest/PreDefinedCodeSnippets.cs
BCSharp: Autoformat. I'm not fighting with it, fixing manually.

PR #339, opened 2026-04-18
Simplify CollectionExtensions in Microsoft.Scripting

BCSharp:
Many became redundant in .NET. The .NET implementations are as good as the ones from `Microsoft.Scripting.dll`, sometimes more optimized, sometimes avoid unnecessary allocations on heap.

The class is internal so no harm done. There is another class `CollectionExtensions` in `Microsoft.Dynamic` that needs to be reviewed too; it is in namespace `Microsoft.Scripting` (?) but sometimes contains simply a copy of functionality from `Microsoft.Scripting.dll`.

### DIFF
--- a/src/core/Microsoft.Dynamic/Interpreter/LightCompiler.cs
+++ b/src/core/Microsoft.Dynamic/Interpreter/LightCompiler.cs
@@ -1162,7 +1162,7 @@ namespace Microsoft.Scripting.Interpreter {
             // force compilation for now for ref types
             // also could be a mutable value type, Delegate.CreateDelegate and MethodInfo.Invoke both can't handle this, we
             // need to generate code.
-            if (!CollectionUtils.TrueForAll(parameters, (p) => !p.ParameterType.IsByRef) ||
+            if (!parameters.All(static p => !p.ParameterType.IsByRef) ||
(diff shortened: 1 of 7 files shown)


slozier (approved):
Thanks! I've been wanting to clean up the Utils for a while. Although I was focused on trying to to obsolete CollectionUtils (which is unfortunately public). I'm guessing the code duplication is because Dynamic (maybe?) can't use the Scripting internals so it was copied to avoid public APIs?

BCSharp:
Oh, yes! Utils are definitely due for a cleanup. It is on my list too, but not the top priority. I still intend to implement async in the DLR, and it is already half April. I went after `CollectionExtensions` because I had noticed some inefficiency there, like creating unnecessary copies of arrays.

`CollectionExtensions` are internal so my guess that's why they are somewhat duplicated. It breaks the DRY rule and I hope I can get rid of them in Scripting altogether. There are other ways to deal with it, but I think this will be the best way.

For the public methods, I thought we could mark them obsolete (with a message pointing to the alternative) and increase the minor DLR version number, since the change is visible on the API level. Then, maybe one day, after some time, increase the major version number and get rid of them for good.

PR #340, opened 2026-04-18
Refactor ToReadOnlyCollection in Microsoft.Scripting

BCSharp:
Follow-up on #339. In the end I did not manage to eliminate `CollectionExtensions` from `Microsoft.Scripting`. There is only one place that uses it, but it receives `IEnumerable` from the .NET runtime, which, in turn, may receive it from the user code, so there is no way of knowing what will be processed. The existing code is not simple. To avoid repeating myself in `Microsoft.Dynamic` I am now really tempted to make it public, since it may be useful to a wider range of (user) code. In `Microsoft.Dynamic` it is being used in a handful of places. It is even being defined (again) in .NET, in `System.Dynamic.Utils`, but not public.

Another options are: make `Microsoft.Dynamic` a friend of `Microsoft.Scripting`, or use `UnsafeAssessorAttribute` (.NET only). I do not consider reflection; I'd rather break DRY.

### DIFF
--- a/src/core/Microsoft.Scripting/Utils/CollectionExtensions.cs
+++ b/src/core/Microsoft.Scripting/Utils/CollectionExtensions.cs
@@ -5,8 +5,12 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
+using System.Linq;
+using System.Runtime.CompilerServices;
 
 namespace Microsoft.Scripting.Utils {
+
+    /// <seealso href="https://github.com/dotnet/runtime/blob/main/src/libraries/System.Linq.Expressions/src/System/Dynamic/Utils/CollectionExtensions.cs"/>
     internal static class CollectionExtensions {
         /// <summary>
         /// Materlializes the provided enumerable if needed and wraps it in a ReadOnlyCollection{T}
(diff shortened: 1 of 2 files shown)


src/core/Microsoft.Scripting/Utils/CollectionExtensions.cs
```diff
@@ -5,43 +5,46 @@
...
-                T[] array = new T[count];
-                collection.CopyTo(array, 0);
-                return new ReadOnlyCollection<T>(array);
-            }
 
-            // ToArray trims the excess space and speeds up access
-            return new ReadOnlyCollection<T>(System.Linq.Enumerable.ToArray(enumerable));
+#if !NET8_0_OR_GREATER
+        // Emulates ReadOnlyCollection<T>.Empty form .NET 8.0+ with what is available in .NET Framework et al.
+        extension<T>(ReadOnlyCollection<T>) {
```
slozier, line 39: Neat. I keep learning new stuff. I really should brush up on the "latest" C# features!

slozier (approved):
Thanks! Looks good to me.

I'm still opposed to adding anything public, especially when it's some generic helper method that's not really domain specific.

I think I'm probably ok with exposing internals since the two assemblies are closely related (although it seems a bit heavy handed if this is the only use case). Another approach that comes to mind is to include the source from the Scripting folder using csproj?

BCSharp:
> I'm still opposed to adding anything public, especially when it's some generic helper method that's not really domain specific.

Yeah, _tempted_ doesn't mean _convinced_. I do not see any good solution right now. Referring to the source file is an interesting idea, but I it doesn't feel good to just grab the file right out of the other's DLL directory. It is not that visible that it's being shared. I'd rather expose it in a directory above like `GlobalUsings.cs`, or create a dedicated location (dir). For this tiny bit, it's not worth the extra complexity, but there is more stuff in `Microsoft.Dynamic` that declares `Microsoft.Scripting` namespace, and that needs to be reviewed, so worth to keeping in mind. The friend assembly is an option I prefer to use for testing, not internal dependencies, so that's that.

Depending on the amount of "shared" stuff after the review of Utils, this is my order of preference (as of now):
1. Small bits (like so far): just copy the code
2. Medium stuff: source file sharing
3. Significant number: friend assembly

BCSharp:
One more thing about `ReadOnlyCollection` I forgot to mention: the .NET version is slightly different because it is stricter. `ReadOnlyCollection` is read-only, but it doesn't mean it is immutable; it's just a wrapper around the `IList` parameter. It is also possible to create a subclass that accesses protected `Items`. Those collections can originate from user code — as `IEnumerable` so who knows what user code does with them. It could be a bug or it could be malicious. The DLR does copy the enumerable that the user provides **except** when it is already `ReadOnlyCollection`. .NET always copies what it gets from the user code, and probably doesn't copy if it comes from within itself using `TrueReadOnlyCollection` — which is internal, thus inaccessible to us. The DLR does not copy what .NET provides if it is a `TrueReadOnlyCollection`, because it is simply a sealed version (subclass) of `ReadOnlyCollection`, but it leads to the caveat above.

All in all, I think the DLR strikes the right balance here between being waterproof and efficient. Considerations may be different at the .NET level.

PR #341, opened 2026-04-19
Upgrade test framework to NUnit 4.5

BCSharp:
For all four tests in the DLR. NUnit 4 has better support for async methods; this may become handy.

Some `HostingTest` assertions have been changed from the classic `Assert.AreEqual` to the new constraint-based `Assert.That` because, although `Assert.AreEqual` is available, it is an extension method now, and extension methods do not work well with `dynamic` arguments. Perhaps it would be neat to change all assertions to the modern form.


slozier (approved):
Thanks. Looks good to me. I don't recall why we initially went with nunuitlite rather than nunit.

PR #342, opened 2026-04-19
Add TFM net9.0

BCSharp:
Having target `net9.0` allows for better testing of features available from .NET 9.0 onwards.

### DIFF
--- a/src/core/Microsoft.Dynamic/Microsoft.Dynamic.csproj
+++ b/src/core/Microsoft.Dynamic/Microsoft.Dynamic.csproj
@@ -1,15 +1,15 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net462;netstandard2.0;net8.0;net10.0</TargetFrameworks>
+    <TargetFrameworks>net462;netstandard2.0;net8.0;net9.0;net10.0</TargetFrameworks>
     <RootNamespace>Microsoft.Scripting</RootNamespace>
     <BaseAddress>859832320</BaseAddress>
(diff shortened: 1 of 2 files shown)


slozier (commented):
Works for me. Note that this might add the net9.0 target to the nuget package. Might need an exclusion if that isn't intended.

BCSharp:
In `eng/package/nuget/DynamicLanguageRuntime.nuspec` I see:

```xml
  <files>
    <file src="**\*.dll" target="lib" exclude="net9.0*\*" />
    <file src="**\*.pdb" target="lib" exclude="net9.0*\*" />
    <file src="**\*.xml" target="lib" exclude="net9.0*\*" />
    <file src="README.md;LICENSE" />
  </files>
```

Doesn't it mean .NET 9.0 is excluded from NuGet?

Whether it should be excluded or included, it is up to you — you organize the releases. At this moment the only benefit from .NET 9 is the fast `Lock` class. Also I wonder whether you will make another release before .NET 9 reaches end-of-support 😄

slozier:
> Doesn't it mean .NET 9.0 is excluded from NuGet?

Wow... I swear I tried to find this this morning. 😢 Guess that's what I get for trying to review stuff using the github app on my phone. 😆  Yeah the `exclude` means it'll get excluded from nuget so all good!

> Whether it should be excluded or included, it is up to you — you organize the releases. At this moment the only benefit from .NET 9 is the fast `Lock` class. Also I wonder whether you will make another release before .NET 9 reaches end-of-support 😄

Looking at their calendar, still 6 months left to go! Let's not lose hope just yet. 😃 Although .NET 8 will also reach end of support at that point. 😞

PR #343, opened 2026-04-20
Run HostingTest on all frameworks

BCSharp:
On .NET(all versions) this means that remoting tests are not compiled — the API is not available. I used the feature compilation symbol like in core DLR, hence the new `Directory.Build.props`.

On Mono, CAS compiles but is not supported. Surprisingly, remoting works.


slozier (approved):
Thanks!

PR #344, opened 2026-04-21
Refactor ReadConfiguration tests to use array syntax for language setups

BCSharp:
I know that this form of array initialization is very popular with some people... 😄 (I must admit I like it too).


slozier (approved):
So much nicer. 😸

PR #348, opened 2026-04-24
Use Meziantou.Polyfill

### LINKED ISSUE #346: Consider using Microsoft.Bcl.Memory
Only on .NET Framework/.NET Standard.

It will implicitly pull in the following additional packages:
* [System.Memory](https://www.nuget.org/packages/System.Memory/) and its dependencies — it is already being used by IronPython
* [System.ValueTuple](https://www.nuget.org/packages/System.ValueTuple/) — only on .NET Framework 4.6.2

The benefit of `Span` and `Memory` is obvious, and for IronPython it does not add extra dependency. 

I missed `ValueTuple` in IronPython on a numerous occasions. I expect the same when modernizing the DLR.

[Microsoft.Bcl.Memory](https://www.nuget.org/packages/Microsoft.Bcl.Memory/) itself provides `System.Range` and `System.Index`, which admittedly we could provide ourselves to enable the new compiler syntax without creating extra dependencies, but then we have the code duplication issue (https://github.com/IronLanguages/dlr/pull/340#issuecomment-4276455513) if we want to be able to use them in various places.

All in all, it will help in modernizing and optimizing the code while keeping it runnable on older frameworks.

Practically, it is conceivable that applications using the DLR or IronPython on older frameworks already pull in those libraries.

This can be done after release 3.4.3 of IronPython, thus wrapping up the current status quo of the DLR.

BCSharp:
By default, _Meizantou.Polyfill_ generates everything that it can that is missing on the target framework, regardless whether it is being used or not. It also not being trimmed away during linking. This is the same as _PolySharp_, but because _Meizantou.Polyfill_ contains much more polyfills, it is significantly more generated code. By my estimates on `net462`, it is ca. ~100K vs. 10K for _PolyFill_. This may be OK for tests, but I regard it as excessive for the production DLLs.

A way to avoid it is to define an allow-list. Features (types, properties, etc.) that are needed and used are listed in the `.csproj` file, others are not generated.

This PR contains only the reference to _Meizantou.Polyfill_ with no polyfils included. I set it up this way so I can work on several things in parallel that depend on polyfills. Even then, on `net462` there are some (empty) types generated, which takes ca. 1K. I think it may be a deficiency in the library that can be remedied, and I may open an issue there, or even submit a PR, if I have time.

Of course all this size increase issues are only relevant on older targets that are missing the newer types; on .NET 10 nothing is being generated so there is no size increase beyond that 1K.

### DIFF
--- a/src/core/Microsoft.Dynamic/Microsoft.Dynamic.csproj
+++ b/src/core/Microsoft.Dynamic/Microsoft.Dynamic.csproj
@@ -6,6 +6,10 @@
     <BaseAddress>859832320</BaseAddress>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <EmitCompilerGeneratedFiles>True</EmitCompilerGeneratedFiles>
+    <MeziantouPolyfill_IncludedPolyfills>
+      None
+    </MeziantouPolyfill_IncludedPolyfills>
   </PropertyGroup>
 
   <PropertyGroup Condition=" '$(TargetFramework)' == 'net9.0' OR '$(TargetFramework)' == 'net10.0' ">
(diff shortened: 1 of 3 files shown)


slozier (approved):
Yes the possibility of size increase did cross my mind but I figured we could filter things to only include what we need if it was an issue. Hopefully I'll find a bit of time to play with it on the IronPython side of things this evening.

PR #351, opened 2026-04-27
Enable nullable annotations in language types in Scripting

BCSharp:
Activates creation of additional polyfills `MaybeNullAttribute` and `NotNullWhenAttribute`.

### DIFF
--- a/src/core/Microsoft.Scripting/Runtime/ScopeStorage.cs
+++ b/src/core/Microsoft.Scripting/Runtime/ScopeStorage.cs
@@ -2,8 +2,11 @@
 // The .NET Foundation licenses this file to you under the Apache 2.0 License.
 // See the LICENSE file in the project root for more information.
 
+#nullable enable
+
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.Dynamic;
 using System.Linq.Expressions;
 using System.Threading;
@@ -30,8 +33,8 @@ namespace Microsoft.Scripting {
         ///
         /// If the named value is not present an InvalidOperationException is raised.
         /// </summary>
-        public dynamic GetValue(string name, bool ignoreCase) {
-            if (GetScopeVariable(name, ignoreCase).TryGetValue(out object res)) {
+        public dynamic? GetValue(string name, bool ignoreCase) {
+            if (GetScopeVariable(name, ignoreCase).TryGetValue(out object? res)) {
                 return res;
             }
             throw new KeyNotFoundException("no value");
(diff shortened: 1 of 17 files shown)


src/core/Microsoft.Scripting/Runtime/SharedIO.cs
```diff
@@ -69,27 +72,27 @@ public override void SetLength(long value) {
...
+        private TextWriter? _outputWriter;
+        private TextWriter? _errorWriter;
 
-        private Encoding _inputEncoding;
+        private Encoding? _inputEncoding;
 
-        public Stream InputStream { get { InitializeInput(); return _inputStream; } }
-        public Stream OutputStream { get { InitializeOutput(); return _outputStream; } }
-        public Stream ErrorStream { get { InitializeErrorOutput(); return _errorStream; } }
+        public Stream InputStream { get { InitializeInput(); return _inputStream!; } }
```
slozier, line 85: Maybe this change is slated for later, but you could use `[MemberNotNull(nameof(_inputStream), nameof(_inputReader), nameof(_inputEncoding))]` on `InitializeInput` ans friends.

src/core/Microsoft.Scripting/Runtime/ScopeStorage.cs
```diff
@@ -288,12 +292,12 @@ bool HasValue {
         /// Atempts to get the value. If a value is assigned it returns true otherwise
         /// it returns false.
         /// </summary>
-        bool TryGetValue(out dynamic value);
+        bool TryGetValue(out object? value);
```
slozier, line 295: I don't really know what the consequences of changing from `dynamic` to `object` are. If it wasn't an accidental change I assume you've thought it through and have a good reason for it. :smile:

BCSharp: I am not sure about this one. The reason I changed it to `object?` is because `SetValue` below accepts `object?`. It seems that the compiler is totally happy with assignment of `object` to `dynamic` without casting, as well as `dynamic` to `object` without casting. So it treats them equivalently? 

One can argue that `dynamic` has more capabilities that `object` so perhaps it should be treated as a specialized `object` (as if a subtype). We can decide on this as a rule — in such case `dynamic?` is more appropiate. If so, there are possibly more places where I need to change things (including future planned PRs).

In `TryGetValue` the type is `out`, and in `SetValue` it is `in`, so it is as if co-variant vs. variant typing, but because the type is actually not changed, it felt weird to set an `object` into the scope, just to retrieve it later magically as `dynamic`.

I need to sleep on this.

src/core/Microsoft.Scripting/Runtime/SharedIO.cs
BCSharp: I thought about it, but to mark those member fields `MemberNotNull`, `InitializeInput` (and friends) would have to change to really guarantee to the compiler they are never null on return. This means that the conditions on line 103 and 105 would have to change to `if (_inputStream is null || _inputEncoding is null || _inputReader is null)`. Since the first test (L103) is executed on _every_ property access, it means that every time the test would take 3 times as much as before — just to make the compiler happy about nullability guarantees that are obvious for the human reading the code (and maybe to some AI agents 😄 ). In the end I thought it was not worth the change, and the null-forgiving operator is justified. What is nice is that the whole anaysis can be done locally; it does not require looking up any other source file.

Generally, I want to do changes file-by-file, in file groups, to make the review easier and for my own well-being, so there is nothing to be slated for later. If I submit a file for a review, that is it, if something looks questionable, suspicious, or unclear, fire right away.

src/core/Microsoft.Scripting/Runtime/ScopeStorage.cs
BCSharp: The type conversion is neither clear not intuitive. Intuitively, the the type of object put into the scope should be the same as the type of the object retrieved from it. This means `object`/`object` or `dynamic`/`dynamic`. But there is no reason for the scope to hold onto the `dynamic` type; it doesn't perform any dynamic operations on it. Therefore type `object` seems more appropriate.

From the caller's perspective it changes nothing: you can still declare the retrieved value as `dynamic` if you want to. For instance, this works:

```c#
if (scopeVariable.TryGetValue(out dynamic? variable) {
    ...
}
```

and obtain a variable typed as `dynamic`.

So in the end, I would not make any changes here.

PR #354, opened 2026-05-01
Upgrade to use Meziantou.Polyfill 1.0.120

BCSharp:
This version is more frugal with generation on targets that already have the requested polyfills, like `net10.0`.


slozier (approved):
Looks like you got a few optimizations in there. 👍

PR #355, opened 2026-05-01
Enable nullable annotations in core of Scripting

BCSharp:
New polyfill: `MaybeNullWhenAttribute`

### DIFF
--- a/src/core/Microsoft.Scripting/Hosting/ExceptionOperations.cs
+++ b/src/core/Microsoft.Scripting/Hosting/ExceptionOperations.cs
@@ -48,7 +48,7 @@ namespace Microsoft.Scripting.Hosting {
             var exceptionObj = exception.Unwrap() as Exception;
             ContractUtils.Requires(exceptionObj is not null, nameof(exception), "ObjectHandle must be to Exception object");
 
-            return _context.FormatException(exceptionObj!);
+            return _context.FormatException(exceptionObj);
         }
 
         public void GetExceptionMessage(ObjectHandle exception, out string message, out string errorTypeName) {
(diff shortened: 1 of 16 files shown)


src/core/Microsoft.Scripting/Hosting/ExceptionOperations.cs
```diff
@@ -46,15 +48,15 @@ public string FormatException(ObjectHandle exception) {
             var exceptionObj = exception.Unwrap() as Exception;
             ContractUtils.Requires(exceptionObj is not null, nameof(exception), "ObjectHandle must be to Exception object");
 
-            return _context.FormatException(exceptionObj);
+            return _context.FormatException(exceptionObj!);
```
slozier: I think you could annotate `ContractUtils.Requires` to avoid using the null-forgiving operator:
```cs
public static void Requires([DoesNotReturnIf(false)] bool precondition, string paramName, string message) {}
```

slozier (approved):

src/core/Microsoft.Scripting/Hosting/ExceptionOperations.cs
BCSharp: Good idea. It wasn't working when I first tried but now I see that the whole `ContractUtils` is defined over again in `Microsoft.Dynamic` and is made public. I must have tried that one. The one in `Scripting` is internal.

A question is why it is exported form `Dynamic`, and not from `Scripting`? That would prevent code duplication. But a bigger question is why does the DLR export such stuff at all? It has nothing to do with the DLR functionality.

slozier: Indeed falls into the should obsolete the whole thing category.

PR #356, opened 2026-05-05
Enable nullable annotations in hosting of Scripting

BCSharp:
No new polyfills; maybe it's enough?

I had to revisit `LanguageContext` from the previous PR.


BCSharp:
The lifetime changed back to `null` (from whatever the base class provided), which apparently is a valid value and means _forever_.

slozier (approved):
Thanks!

PR #357, opened 2026-05-06
Remove nullable annotations directives from Scripting

BCSharp:
Since all files in `Microsoft.Scripting` are annotated, it enables nullable annotations at the whole project level, rather than individually at the beginning of each file.

Two generated files still have directive `#nullable enable` because nullability project setting does not apply to _generated_ files. But something has to be done with them:
- one is generated and checked during tests by IronPython — this is inappropiate for the DLR repo to depend on the IronPython repo;
- the other is "self contained" but being generated by IronRuby, clearly now defunct.

Perhaps best would be to use Roslyn code generators.

There are much more such cases in `Microsoft.Dynamic`.

### DIFF
--- a/src/core/Microsoft.Scripting/ArgumentTypeException.cs
+++ b/src/core/Microsoft.Scripting/ArgumentTypeException.cs
@@ -2,8 +2,6 @@
 // The .NET Foundation licenses this file to you under the Apache 2.0 License.
 // See the LICENSE file in the project root for more information.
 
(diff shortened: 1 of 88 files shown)


slozier (approved):

slozier:
> Perhaps best would be to use Roslyn code generators.

Agreed on this, been wanting to look into those for a while, but never find the time. I don't like the dependency on ironpython.

PR #360, opened 2026-05-12
Generate polyfills for CA2263

BCSharp:
Also updated Meziantou.Polyfill to 1.0.127, which has our optimized `string.IndexOf`


slozier (approved):
Thanks! Surprised there aren't more of these. Does the analyzer skip `.Generated` files on its own or is it some setting we have? For example `CallInstruction.Generated.cs` is full of these.

BCSharp:
Generated files are normally [excluded from code analysis](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/configuration-options#exclude-generated-code).

In this case I did not care about the generated code, only about human-written code, so I could remove the warning silenced from the project file. Code generation requires a review on its own.

PR #363, opened 2026-06-17
Prevent transforming of async expressions

BCSharp:
The current debug instrumentation machinery has no understanding of the new `AsyncExpression` and `AsyncEnumerableExpression` and hard-fails trying to transform them.

This PR prevents the existing mechanism descending into those expressions. It prevents crashes when instrumentation is enabled, but it also leaves async expressions un-debuggable — for now. The debug instrumentation of async expressions deserves it own PR.

When used in IronPython, it will "resolve" IronLanguages/ironpython3#2057. "Resolve" in quotes because the crash is gone, but instrumentation is still not implemented (this is already true for .NET-based coroutines, though — i.e. no crash, no instrumentation).


slozier (approved):
Thanks!

